PR #12, opened 2019-07-23
Make trace headers case insensitive

DarcyRaynerDD:
*Note: Please remember to review the [contribution guidelines](https://github.com/DataDog/datadog-lambda-layer-js/blob/master/CONTRIBUTING.md)
if you have not yet done so.*

### What does this PR do?

Reads trace headers with case insensitivity.

### Motivation

We were having an issue with some frameworks which would force mixed casing.

### Testing Guidelines

How did you test this pull request?

### Additional Notes

Anything else we should know when reviewing?

### DIFF
--- a/src/trace/context.ts
+++ b/src/trace/context.ts
@@ -64,13 +64,11 @@ export function readTraceFromEvent(event: any): TraceContext | undefined {
     return;
   }
 
-  const lowerCaseHeaders = Object.keys(headers).reduce(
-    (prev, header) => {
-      prev[header.toLocaleLowerCase()] = headers[header];
-      return prev;
-    },
-    {} as { [key: string]: string },
-  );
+  const lowerCaseHeaders: { [key: string]: string } = {};
+
+  for (const key of Object.keys(headers)) {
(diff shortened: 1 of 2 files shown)


src/trace/context.ts
```diff
@@ -58,15 +63,24 @@ export function readTraceFromEvent(event: any): TraceContext | undefined {
   if (typeof headers !== "object") {
     return;
   }
-  const traceID = headers[traceIDHeader];
+
+  const lowerCaseHeaders = Object.keys(headers).reduce(
```
tianchu: Just a nits/personal preference, feel free to ignore.

I feel `map` makes it slightly easier to read than `reduce`, which normally reduces an object/array to a single value?

```javascript
  const lowerCaseHeaders = {};
  Object.keys(headers).map(key => {lowerCaseHeaders[key.toLocaleLowerCase()]= headers[key]})
```

DarcyRaynerDD: Since it isn't returning anything in the map function, it's better to just use a for loop. I've updated this.

PR #20, opened 2019-08-13
Misspelling in README.md

a7IadXE1c9:
Should be `module.exports` instead `exports.exports`


DarcyRaynerDD:
@a7IadXE1c9 Thats a good catch, thanks for the PR.

PR #21, opened 2019-09-05
Experimental Integration with dd-trace

DarcyRaynerDD:
### What does this PR do?

Adds experimental integration with the dd-trace library, by exporting trace metadata to logs.

### DIFF
--- a/src/trace/datadog-tracer.ts
+++ b/src/trace/datadog-tracer.ts
@@ -1,5 +1,5 @@
 import Tracer from "dd-trace";
-import { logDebug } from "utils";
+import { logDebug } from "../utils";
 
 const AWS_LAMBDA_FUNCTION_NAME = "AWS_LAMBDA_FUNCTION_NAME";
 
--- a/src/trace/datadog-tracer.ts
+++ b/src/trace/datadog-tracer.ts
@@ -1,9 +1,13 @@
 import Tracer from "dd-trace";
+import { logDebug } from "utils";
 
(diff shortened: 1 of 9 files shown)


src/trace/datadog-tracer.ts
```diff
@@ -0,0 +1,14 @@
+import Tracer from "dd-trace";
+
+const AWS_LAMBDA_FUNCTION_NAME = "AWS_LAMBDA_FUNCTION_NAME";
+
+export function initDatadogTracer() {
+  const functionName = process.env[AWS_LAMBDA_FUNCTION_NAME];
+  const service = functionName ? `${functionName}` : "lambda";
```
tianchu: Is it a surprise if we need to default service name to `lambda` becasue function name is unavailable? If yes, we probably should log a warning to let user know? Otherwise, users will be super confused when this happens.

src/trace/datadog-tracer.ts
```diff
@@ -0,0 +1,14 @@
...
+  Tracer.init({
+    experimental: {
+      exporter: "log",
+    },
+    service,
+  });
+}
```
tianchu: Can we add a debug log for tracer init complete to help make sure tracer is indeed initialized when customers having issues with tracing not working?

src/trace/datadog-tracer.ts
DarcyRaynerDD: Mostly for where the user is running in a test environment where that variable might not be defined. I'll log an error in that case.

PR #26, opened 2019-09-20
Darcy.rayner/trigger info

DarcyRaynerDD:
### What does this PR do?

Detects the event source/trigger of the event, and adds it as a tag to the trace.
Screen Shot 2019-09-20 at 3 28 49 PM

### Motivation

Will eventually enable filtering by trigger/event source type on the serverless page.


src/trace/listener.ts
```diff
@@ -47,6 +50,7 @@ export class TraceListener {
     if (this.context) {
       options.tags = {
         cold_start: this.coldstart,
+        event_source: this.eventSource,
```
sfirrin: What would you think about calling this something like `event_service` to leave open the possibility of making a tag like `event_resource` in the future that would be the ARN of the resource that triggered the function?

DarcyRaynerDD: Yeah that makes sense, or maybe "trigger_type", since it seems related to the concept of a trigger in the AWS Lambda UI, but triggers don't exist as a concept in the underlying infrastructure/resources.

sfirrin: yeah "trigger_type" would make sense to me. @pinkerton want to weigh in?

PR #28, opened 2019-10-11
Add enhanced lambda metrics

sfirrin:
### What does this PR do?

Increments the `aws.lambda.enhanced.invocations` distribution metric on each function call and increments `aws.lambda.enhanced.errors` when the Lambda handler throws an exception.

### Testing Guidelines

Unit tests for the metrics, currently adding the updated layer to functions in the demo account.

### DIFF
--- a/src/metrics/enhanced-metrics.ts
+++ b/src/metrics/enhanced-metrics.ts
@@ -5,22 +5,12 @@ import { getColdStartTag } from "../utils/cold-start";
 
 const ENHANCED_LAMBDA_METRICS_NAMESPACE = "aws.lambda.enhanced";
 
-function areEnhancedMetricsEnabled() {
-  return getEnvValue("DD_ENHANCED_METRICS", "false").toLowerCase() === "true";
-}
-
 export function incrementInvocationsMetric(functionARN: string): void {
-  if (!areEnhancedMetricsEnabled()) {
-    return;
-  }
   const tags = [...parseTagsFromARN(functionARN), getColdStartTag()];
   sendDistributionMetric(`${ENHANCED_LAMBDA_METRICS_NAMESPACE}.invocations`, 1, ...tags);
 }
 
 export function incrementErrorsMetric(functionARN: string): void {
-  if (!areEnhancedMetricsEnabled()) {
-    return;
-  }
   const tags = [...parseTagsFromARN(functionARN), getColdStartTag()];
   sendDistributionMetric(`${ENHANCED_LAMBDA_METRICS_NAMESPACE}.errors`, 1, ...tags);
 }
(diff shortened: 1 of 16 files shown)


src/utils/handler.ts
```diff
@@ -18,6 +19,7 @@ export function wrap<TEvent, TResult>(
   return async (event: TEvent, context: Context) => {
     try {
       await onStart(event, context);
+      incrementInvocationsMetric(context.invokedFunctionArn);
```
DarcyRaynerDD: The incrementInvocationsMetric/incrementErrorsMetric  logic should probably be  moved inside the onStart/onComplete handlers. The idea behind this class is that it's a utility that could be moved out of dd-lambda-js, and potentially be used to wrap one handler multiple times. onComplete can be modified to accept an result or error object.

src/utils/cold-start.ts
```diff
@@ -0,0 +1,27 @@
+let _didFunctionColdStart = true;
```
DarcyRaynerDD: We should reuse this logic in the trace handler, since that tracks cold starts as well: https://github.com/DataDog/datadog-lambda-layer-js/blob/2e6ee435fd009896384184cce3dfa40f68b75fe5/src/trace/listener.ts#L19

src/utils/cold-start.ts
```diff
@@ -0,0 +1,27 @@
...
+
+let _isColdStartSet = false;
```
DarcyRaynerDD: This is probably inconsistent with how we write js in other repos, but I don't think this repo uses underscores for global variables anywhere else.

src/utils/handler.ts
DarcyRaynerDD: Also, do you think we should make this an opt-in config settings? I'm a little worried about customer's accidentally adding synchronous metric calls to their lambdas.

src/metrics/enhanced-metrics.ts
```diff
@@ -0,0 +1,26 @@
+import { getEnvValue, sendDistributionMetric } from "../index";
+
+import { parseTagsFromARN } from "../utils/arn";
+import { getColdStartTag } from "../utils/cold-start";
+
+const ENHANCED_LAMBDA_METRICS_NAMESPACE = "aws.lambda.enhanced";
+
+function areEnhancedMetricsEnabled() {
+  return getEnvValue("DD_ENHANCED_METRICS", "false").toLowerCase() === "true";
```
DarcyRaynerDD: Nice work with most of these changes. One small thing, the other config options are usually exposed both as an environment variable and a [config property](https://github.com/DataDog/datadog-lambda-layer-js/blob/2e6ee435fd009896384184cce3dfa40f68b75fe5/src/index.ts#L30) which we resolve in index.ts . Would you be able to move this logic there?

Issue #32, opened 2019-11-08
Timeout when writing function that never returns


DarcyRaynerDD:
## Expected Behavior
When writing a handler that never returns a value, using either the async await method or the callback approach, the wrapper will block the function from completing. This doesn't happen when the wrapper isn't used.

## Actual Behavior
Function times out

```js
exports.handler = datadog(function (event, context) { });
// Timeout occurs
```
As a work around, prepending async to the function fixes this
```
exports.handler = datadog(async function (event, context) { });
// No timeout
```

## Specifications

  - Datadog Lambda Layer version: 5 +
  - Node version: 10

DarcyRaynerDD:
This was fixed in pr #33

PR #33, opened 2019-11-11
Darcy.rayner/fix timeout no return

DarcyRaynerDD:
### What does this PR do?

Fixes the behavior of the handler wrapper to match the lambda node runtime in the following cases:
- Handlers that never return a value will no longer block until timeout
- Handlers that return a non promise value will have their return values ignored
- Handlers the use the callback, and return a promise, will return with the first to complete

### Motivation
Issue #32 

### Testing Guidelines

Added test cases, and manual testing.

### DIFF
--- a/src/metrics/listener.ts
+++ b/src/metrics/listener.ts
@@ -1,6 +1,6 @@
 import { promisify } from "util";
 
-import { logError } from "../utils";
+import { logError, logDebug } from "../utils";
 import { APIClient } from "./api";
 import { KMSService } from "./kms-service";
 import { Distribution } from "./model";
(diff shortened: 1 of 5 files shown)


src/metrics/listener.ts
```diff
@@ -124,8 +127,13 @@ export class MetricsListener {
       } catch (error) {
         logError("couldn't decrypt kms api key", { innerError: error });
       }
-    } else {
-      logError("api key not configured");
+    } else if (!config.logForwarding) {
+      const errorMessage = "api key not configured, see https://dtdg.co/sls-node-metrics";
+      if (config.logForwarding) {
+        logDebug(errorMessage);
```
brianhyder: It doesn't appear that this will ever execute because the `if` condition is the opposite of the condition above it.

DarcyRaynerDD: Good catch @brianhyder

PR #48, opened 2020-01-21
Fix timestamps for async metrics

sfirrin:
### What does this PR do?

Fixes the listener to log async metrics with Unix second timestamps rather than the milliseconds that it's currently using. The DD API accepts seconds only so this is causing issues with async metrics when submitted in the log forwarder.

### Testing Guidelines

Deployed and tested the layer in staging env and added new unit tests for the metric log generating code.

### DIFF
--- a/src/metrics/build-metric-log.ts
+++ b/src/metrics/build-metric-log.ts
@@ -1,0 +1,10 @@
+// Builds the string representation of the metric that will be written to logs
+export function buildMetricLog(name: string, value: number, tags: string[]) {
+  return `${JSON.stringify({
+    // Date.now() returns Unix time in milliseconds, we convert to seconds for DD API submission
+    e: Date.now() / 1000,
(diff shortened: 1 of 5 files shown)


src/metrics/build-metric-log.ts
```diff
@@ -0,0 +1,10 @@
+// Builds the string representation of the metric that will be written to logs
+export function buildMetricLog(name: string, value: number, tags: string[]) {
+  return `${JSON.stringify({
+    // Date.now() returns Unix time in milliseconds, we convert to seconds for DD API submission
+    e: Date.now() / 1000,
```
DarcyRaynerDD: Should this be rounded?

sfirrin: the decimals don't seem to make a difference, in testing the metrics are ingested ok with the decimal timestamps

PR #50, opened 2020-02-04
Enable enhanced metrics by default and write enhanced metrics to logs

sfirrin:
### What does this PR do?

* Enables enhanced metrics by default (they can be disabled with DD_ENHANCED_METRICS=false)
* If DD_ENHANCED_METRICS is enabled, always writes enhanced metrics to stdout
* recognizes the Node v12 runtime and adds that tag to enhanced metrics

### Motivation

Making enhanced metrics GA

### Testing Guidelines

Updated unit tests, tested the new layer version in our demo environment


src/metrics/enhanced-metrics.ts
```diff
@@ -52,10 +58,20 @@ export function getEnhancedMetricTags(context: Context): string[] {
   return tags;
 }
 
+/**
+ * Increments the specified enhanced metric, applying all relevant tags
+ * @param context object passed to invocation by AWS
+ * @param metricName name of the enhanced metric without namespace prefix, i.e. "invocations" or "errors"
+ */
+function incrementEnhancedMetric(metricName: string, context: Context) {
```
agocs: nice

src/metrics/enhanced-metrics.ts
```diff
@@ -30,6 +32,10 @@ export function getRuntimeTag(): string | null {
     processVersionTagString = RuntimeTagValues.Node10;
   }
 
+  if (processVersion.startsWith("v12")) {
```
DarcyRaynerDD: That's a good bug catch

PR #57, opened 2020-03-06
Only wrap the handler once

tianchu:
### What does this PR do?

When user uses the datadog serverless plugin and also have the handler wrapped, it will result in the handler being wrapped twice and cause enhanced metrics being emitted twice on each invocation (there are probably other side effects). The result looks like

```js
module.exports.handler = datadog(datadog(handler));
```

Fix this by wrapping the handler only once, while `forceWrap` allow wrapping to be forced, which is useful for unit tests.


src/index.ts
```diff
@@ -65,6 +68,7 @@ let currentTraceListener: TraceListener | undefined;
  */
 export function datadog<TEvent, TResult>(
   handler: Handler<TEvent, TResult>,
+  forceWrap = false,
```
DarcyRaynerDD: Can we add forceWrap to the config object instead? Changing the signature of this handler means any customer using the config parameter will have to update their code. It also means we have to go in and update the serverless-plugin to use the new property.

src/index.ts
```diff
@@ -73,6 +81,12 @@ export function datadog<TEvent, TResult>(
   const traceListener = new TraceListener(finalConfig, handlerName);
   const listeners = [metricsListener, traceListener];
 
+  // Only wrap the handler once unless forced
```
DarcyRaynerDD: This is my bad, but I just realized this might be kind of fragile, if we have two handlers in a single file, I think this will break.

eg.
```
const firstFunction = datadatog(() => {..});
const secondFunction = datadatog(() => {..}); // This won't get wrapped, customer would have to use forceWrap
```

I think you might have to do some fancy logic inside the bottom lambdas

PR #59, opened 2020-03-12
Make getEnhancedMetricTags() works for local invocation

jeromemacias:
### What does this PR do?

When we run local invocation using Serverless framework, the `invokedFunctionArn` context value is undefined.

### Motivation

I really want to test all my lambda in local mode, even if it's very difficult, it's so important to me.

### Testing Guidelines

Run a lambda using invoke command locally of the serverless framework with datadog wrapper.


DarcyRaynerDD:
Hi @jeromemacias, thanks for the PR. We are going to roll this into the next release.

PR #60, opened 2020-03-13
Upgrade dependencies, drop node 8 support

DarcyRaynerDD:
### What does this PR do?

Upgrades dependencies and drops node 8 support, (this is now fully deprecated by AWS).

### Motivation

Latest version of dd-trace has new aws sdk integration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ arn:aws:lambda:<AWS_REGION>:464622532012:layer:Datadog-Node12-x:<VERSION>
 # OR
 arn:aws:lambda:<AWS_REGION>:464622532012:layer:Datadog-Node10-x:<VERSION>
 # OR
-arn:aws:lambda:<AWS_REGION>:464622532012:layer:Datadog-Node8-10:<VERSION>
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -31,7 +31,7 @@ arn:aws:lambda:<AWS_REGION>:464622532012:layer:Datadog-Node12-x:<VERSION>
 # OR
 arn:aws:lambda:<AWS_REGION>:464622532012:layer:Datadog-Node10-x:<VERSION>
 # OR
-arn:aws:lambda:<AWS_REGION>:464622532012:layer:Datadog-Node8-10:<VERSION>
+arn:aws:lambda:<AWS_REGION>:464622532012:layer:Datadog-Node8-10:<VERSION> # (Deprecated)
```
agocs: Given that Node 8 functions can't be updated (and therefore no one new can install the Node 8 layer), does it make sense to just delete this?

DarcyRaynerDD: I think we should leave it there, just to make it clear to anyone upgrading there old lambdas to know what to search for. Maybe we can remove it in a few months?

PR #62, opened 2020-03-21
Be able to provide custom logger to replace console usage

jeromemacias:
### What does this PR do?

Add possibility to provide a custom logger object instead of `console` usage for `logError` and `logDebug`

### Motivation

I use Pino as logger in my lambda, so I wanted to format error and debug log from datadog lambda wrapper as my other logs

### Testing Guidelines

Pass a custom logger option and set log level as debug to see logs

### Additional Notes

~I will add tests and documentation if new feature is accepted~
Tests and documentation added

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -8,7 +8,7 @@ import {
   MetricsListener,
 } from "./metrics";
 import { TraceConfig, TraceHeaders, TraceListener } from "./trace";
-import { logError, LogLevel, setColdStart, setLogLevel, setLogger, wrap } from "./utils";
+import { logError, LogLevel, Logger, setColdStart, setLogLevel, setLogger, wrap } from "./utils";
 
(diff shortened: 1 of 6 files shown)


src/index.ts
```diff
@@ -33,6 +33,10 @@ interface GlobalConfig {
    * @default false
    */
   forceWrap: boolean;
+  /**
+   * Custom logger.
+   */
+  logger?: any;
```
DarcyRaynerDD: Would you be able to create an interface for this?

jeromemacias:
@DarcyRaynerDD Thanks for the review, I will finish this PR asap

src/index.ts
jeromemacias: Done

jeromemacias:
@DarcyRaynerDD Tests and documentation added ;)

PR #65, opened 2020-03-27
Add documentation and example for adding request ID to logs.

agocs:
### What does this PR do?

This adds documentation around how to add the AWS request ID to logs when using a custom logger, in order to achieve trace-log correlation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,37 @@ sendDistributionMetric(
 
 If your Lambda function is associated with a VPC, you need to ensure it has access to the [public internet](https://aws.amazon.com/premiumsupport/knowledge-center/internet-access-lambda-function/).
 
+## Log correlation with other logging libraries
+
+In order to correlate logs emitted by your Lambda with specific invocations, it
+is necessary to add the AWS Request ID to your logs. This is done automatically
+for `console.log()`, but you will have to implement this for other logging libraries.
+
+The AWS Request ID is available in the context that is passed to your lambda handler,
+as `context.awsRequestId`. It should be included in your log line as `@lambda.request_id`.
+
+For example, using the [Pino](https://getpino.io/) logger:
+
+```js
+
+const logger = require('pino')()
+
+exports.handler = async function(event, context) {
+ 
+  //This sets up your request-specific logger to emit logs with the Request ID property.
+  const req_logger = logger.child({ '@lambda.request_id': context.awsRequestId });
+
+  //Carry on with whatever the lambda needs to do
+  const work = do.Work();
+
+  //Write a log messagew
+  req_logger.info("Work complete");
+
+  return work;
+}
+
+```
+
 ## Distributed Tracing
 
 [Distributed tracing](https://docs.datadoghq.com/tracing/guide/distributed_tracing/?tab=nodejs) allows you to propagate a trace context from a service running on a host to a service running on AWS Lambda, and vice versa, so you can see performance end-to-end. Linking is implemented by injecting Datadog trace context into the HTTP request headers.
--- a/README.md
+++ b/README.md
@@ -147,9 +147,8 @@ as `context.awsRequestId`. It should be included in your log line as `@lambda.re
 
 For example, using the [Pino](https://getpino.io/) logger:
 
-```js
-
-const logger = require('pino')()
+```typescript
+const logger = require('pino')();
 
 exports.handler = async function(event, context) {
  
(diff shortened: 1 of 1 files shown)


agocs:
The example is based on https://getpino.io/#/?id=usage , for reference.

README.md
```diff
@@ -136,6 +136,36 @@ sendDistributionMetric(
 
 If your Lambda function is associated with a VPC, you need to ensure it has access to the [public internet](https://aws.amazon.com/premiumsupport/knowledge-center/internet-access-lambda-function/).
```
tianchu: I think this title can be misleading, because this only applies when you are not using datadog tracer.

tianchu: I think we should have a section `Trace & Log Correlation` and break it down by using datadog tracer and using xray tracer.

tianchu: Let's move the description for `DD_LOGS_INJECTION` to this new section, since it's getting a little bit complicated.

agocs: Using the datadog tracer, trace-log correlation should just work, right? Or is there something special they need to do?

README.md
```diff
@@ -136,6 +136,36 @@ sendDistributionMetric(
...
+
+For example, using the [Pino](https://getpino.io/) logger:
+
+```typescript
+const logger = require('pino')();
+
+exports.handler = async function(event, context) {
+ 
+  //This sets up your request-specific logger to emit logs with the Request ID property.
+  const req_logger = logger.child({ '@lambda.request_id': context.awsRequestId });
```
sfirrin:
```suggestion
  const req_logger = logger.child({ 'lambda.request_id': context.awsRequestId });
```

The `@` prefix is used [to query for facets in logs](https://docs.datadoghq.com/logs/search_syntax/#facets-search), so we don't want to add an actual `@` to the log attribute.

README.md
tianchu: Correct, it just works if I'm using datadog tracer. However, reading this section as a user, I wouldn't know that. Even if I'm using the datadog tracer, I still might think I need to follow the instructions in this section.

README.md
```diff
@@ -136,6 +136,36 @@ sendDistributionMetric(
...
 If your Lambda function is associated with a VPC, you need to ensure it has access to the [public internet](https://aws.amazon.com/premiumsupport/knowledge-center/internet-access-lambda-function/).
 
+## Log correlation with other logging libraries
+
+In order to correlate logs emitted by your Lambda with specific invocations, it
+is necessary to add the AWS Request ID to your logs. This is done automatically
+for `console.log()`, but you will have to implement this for other logging libraries.
+
+The AWS Request ID is available in the context that is passed to your lambda handler,
+as `context.awsRequestId`. It should be included in your log line as `@lambda.request_id`.
```
tianchu: Is this `@` sign needed?

README.md
tianchu: I don't think this is a super big deal though, definitely not a blocker for merging this PR, we can always improve the structure later.

agocs: 👍 Updated!

README.md
```diff
@@ -87,11 +87,8 @@ If you set the value of this variable to "true" then the Lambda layer will incre
 
 ### DD_LOGS_INJECTION
 
-By default, the Datadog trace id gets automatically injected into the logs for correlation, if using `console` or a logging library supported for [automatic](https://docs.datadoghq.com/tracing/connect_logs_and_traces/?tab=nodejs#automatic-trace-id-injection)  trace id injection.
-
-See instructions for [manual](https://docs.datadoghq.com/tracing/connect_logs_and_traces/?tab=nodejs#manual-trace-id-injection) trace id injection, if using other logging libraries.
-
-Set the environment variable `DD_LOGS_INJECTION` to `false` to disable this feature.
+Controlls whether or not the request ID is injected into log lines. See [DD_LOGS_INJECTION](#DD_LOGS_INJECTION-environment-variable) under
```
tianchu: @DarcyRaynerDD correct me if i'm wrong. I thought DD_LOG_INJECTION only inject datadog trace id into logs?

agocs: Okay yep, I misread the other section.

README.md
```diff
@@ -136,6 +133,52 @@ sendDistributionMetric(
 
 If your Lambda function is associated with a VPC, you need to ensure it has access to the [public internet](https://aws.amazon.com/premiumsupport/knowledge-center/internet-access-lambda-function/).
 
+## Trace & Log Correlation
+
+### Using the Datadog Tracer
+
+If you are using the [Datadog Tracer](#datadog-tracer-experimental), your log messages
```
DarcyRaynerDD: This isn't quite true. You still need to follow the steps (here)[https://docs.datadoghq.com/tracing/connect_logs_and_traces/nodejs/]

agocs: 👍 Thanks. Fixed.

PR #69, opened 2020-04-09
Lowercases function arn on dd-trace spans

DarcyRaynerDD:
### What does this PR do?

Makes the dd-trace root span use a lowercase only function arn. This guarantees the trace will appear on the serverless traces page.

### DIFF
--- a/src/trace/listener.ts
+++ b/src/trace/listener.ts
@@ -89,9 +89,10 @@ export class TraceListener {
     const options: TraceOptions = {};
     if (this.context) {
       logDebug("Applying lambda context to datadog traces");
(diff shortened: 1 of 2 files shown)


src/trace/listener.ts
```diff
@@ -89,9 +89,10 @@ export class TraceListener {
     const options: TraceOptions = {};
     if (this.context) {
       logDebug("Applying lambda context to datadog traces");
+      const functionArn = (this.context.invokedFunctionArn ?? "").toLowerCase();
       options.tags = {
         cold_start: didFunctionColdStart(),
-        function_arn: this.context.invokedFunctionArn,
+        function_arn: functionArn,
```
tianchu, line 95: Do we need to worry about `-` vs `_` too?

PR #71, opened 2020-04-27
Fix xray-sdk related errors

DarcyRaynerDD:
### What does this PR do?
New version of the xray sdk had significant behaviour changes. Unless initialized, getSegment  and captureFunc would throw an error or log, instead of returning empty. The getSegment call was particularly bad, because it occurred during the patched console.log method, and could generate an infinite loop. This PR:

1. Fixes the console.log issue, by disabling the xray logger temporarily while calling getSegment, catching any errors and also adding a re-entrance check to the patched function. 
1. Removes the captureFunc call used to add datadog metadata to the x-ray trace, and replaces it by calling the daemon directly.

### Motivation

What inspired you to submit this pull request?

### Testing Guidelines

How did you test this pull request?

### Additional Notes

Anything else we should know when reviewing?

### DIFF
--- a/src/trace/context.ts
+++ b/src/trace/context.ts
@@ -118,21 +118,27 @@ export function generateXraySubsegment(key: string, metadata: Record<string, any
 export function sendXraySubsegment(segment: string) {
   const xrayDaemonEnv = process.env[awsXrayDaemonAddressEnvVar];
   if (xrayDaemonEnv === undefined) {
+    logDebug("X-Ray daemon env var not set, not sending sub-segment");
     return;
   }
   const parts = xrayDaemonEnv.split(":");
   if (parts.length <= 1) {
+    logDebug("X-Ray daemon env var has invalid format, not sending sub-segment");
     return;
   }
   const port = parseInt(parts[1], 10);
   const address = parts[0];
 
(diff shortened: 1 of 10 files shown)


src/trace/context.ts
```diff
@@ -69,14 +73,66 @@ export function addTraceContextToXray(traceContext: TraceContext) {
...
+  });
+}
+
+export function sendXraySubsegment(segment: string) {
+  const xrayDaemonEnv = process.env[awsXrayDaemonAddressEnvVar];
+  let port = xrayUDPPort;
+  let address = localHost;
+  if (xrayDaemonEnv !== undefined) {
+    const parts = xrayDaemonEnv.split(":");
+    if (parts.length > 1) {
```
agocs, line 126: Is there any sense in continuing if we can't parse the xrayDaemonEnv? It feels like we should mock that in tests, and fail + debug log if we can't find it in production

DarcyRaynerDD: Yeah, you're right, it probably makes sense to bail in that case. I was thinking I'd just copy what the xray-sdk does, which is to have the defaults. But I guess if it's not set, then we probably aren't in a lambda environment

DarcyRaynerDD: Okay, I've implemented this feedback

PR #74, opened 2020-05-21
Add default 'handler'

DarcyRaynerDD:
### What does this PR do?

Adds a 'handler' file, that can be pointed to by the user without having to change their code. Using this file involves two steps.
1. Set the handler to ''/opt/nodejs/node_modules/datadog-lambda-js/handler.handler"
2. Set 'DD_LAMBDA_HANDLER' to the actual handler


src/index.ts
```diff
@@ -9,18 +9,19 @@ import {
...
-
-const defaultSiteURL = "datadoghq.com";
+export const apiKeyEnvVar = "DD_API_KEY";
+export const apiKeyKMSEnvVar = "DD_KMS_API_KEY";
+export const siteURLEnvVar = "DD_SITE";
+export const logLevelEnvVar = "DD_LOG_LEVEL";
+export const logForwardingEnvVar = "DD_FLUSH_TO_LOG";
+export const logInjectionEnvVar = "DD_LOGS_INJECTION";
+export const enhancedMetricsEnvVar = "DD_ENHANCED_METRICS";
+export const datadogHandlerEnvVar = "DATADOG_USER_HANDLER";
```
tianchu, line 21: How about `DD_LAMBDA_HANDLER`?

shellscape:
This seems... incredibly inconvenient for users who have a multitude of functions setup with a single serverless config. Was that considered when this PR was being considered?

DarcyRaynerDD:
Hi @shellscape, this feature is totally optional. It was designed to cut down the amount of boilerplate required to instrument a large number of functions, so a user doesn't have to make any changes to their code, just their configuration. It's particularly useful when using serverless-plugin-datadog, datadog-ci or our SAM macro. But if you prefer the previous way, where you manually wrap each of your handlers in code, then that is still supported.

PR #79, opened 2020-07-09
Parse function version from function arn

hghotra:
### What does this PR do?

This changes in the PR allow for parsing of the function version from the function arn if there is one. If there is no function version, then a default value of $LATEST is set.

### Motivation

https://datadoghq.atlassian.net/browse/SLS-588

### Testing Guidelines

Manually tested using a locally built version of datadog-lambda with the required changes, and making sure the changes make it all the way to the datadog traces.

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "@types/node": "^13.9.1",
     "@types/promise-retry": "^1.1.3",
     "@types/shimmer": "^1.0.1",
-    "dd-trace": "0.20.3",
+    "dd-trace": "0.23.2",
     "jest": "^25.1.0",
     "nock": "12.0.2",
     "prettier": "^1.19.1",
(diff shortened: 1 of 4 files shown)


src/trace/listener.spec.ts
```diff
@@ -80,6 +80,7 @@ describe("TraceListener", () => {
         tags: {
           cold_start: true,
           function_arn: "arn:aws:lambda:us-east-1:123456789101:function:my-lambda",
+          function_version: "$Latest",
```
tianchu, line 83: Shouldn't it be `$LATEST`?

hghotra: Yes, you're right ..fixed.

package.json
```diff
@@ -22,7 +22,7 @@
     "@types/node": "^13.9.1",
     "@types/promise-retry": "^1.1.3",
     "@types/shimmer": "^1.0.1",
-    "dd-trace": "0.20.3",
+    "dd-trace": "0.23.2",
```
tianchu, line 25: 👍

PR #84, opened 2020-07-28
[SLS-595] Update layer to use new tracing format

hghotra:
### What does this PR do?

Update layer to use new tracing format.

### Motivation

https://datadoghq.atlassian.net/browse/SLS-595

### Testing Guidelines

Tested manually.

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [x] Breaking change

### Check all that apply


src/handler.ts
```diff
@@ -5,7 +5,11 @@ const { load } = require("/var/runtime/UserFunction") as any;
 
 if (getEnvValue("DD_TRACE_ENABLED", "true").toLowerCase() === "true") {
   // tslint:disable-next-line:no-var-requires
-  require("dd-trace").init();
+  require("dd-trace").init({
```
DarcyRaynerDD, line 8: This only applies for the redirected handler. It won't apply if they are using the wrapper manually.

hghotra: If they use the wrapper manually, then the forwarder seems to take care of it.

src/trace/listener.ts
```diff
@@ -111,7 +114,9 @@ export class TraceListener {
       options.childOf = spanContext;
     }
     options.type = "serverless";
-    options.resource = this.handlerName;
+    options.service = "aws.lambda";
+    options.resource = this.context && this.context.functionName;
```
DylanLovesCoffee, line 118: In the case that `this.context` is non-existent/falsy, would `resource` display the falsy value as a string?

hghotra: What should the `resource` be if the `context` is `undefined`? Currently, `context` gets initialized here: https://github.com/DataDog/datadog-lambda-js/blob/b64a50e3316fff7df9d0d17ae56db06180690262/src/trace/listener.ts#L61
After the initialization, if `context` is `undefined`, then resource will get set to `undefined` - maybe we shouldn't set it at all?

DylanLovesCoffee: Yeah I think that's a good idea - if `context` doesn't exist then let's skip assigning a value to `resource` altogether. Seems like we currently do that in our ruby and python layers.

hghotra: Fixed.

PR #96, opened 2020-08-19
Add script to check AWS regions missing lambda layers

hghotra:
### What does this PR do?

Add script to check AWS regions missing lambda layers
Add script to add lambda layers to new region

### Motivation

https://datadoghq.atlassian.net/browse/SLS-675

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


scripts/list_layers.sh
```diff
@@ -11,19 +11,20 @@
 set -e
 
 LAYER_NAMES=("Datadog-Node10-x" "Datadog-Node12-x")
-AVAILABLE_REGIONS=(us-east-2 us-east-1 us-west-1 us-west-2 ap-east-1 ap-south-1 ap-northeast-2 ap-southeast-1 ap-southeast-2 ap-northeast-1 ca-central-1 eu-north-1 eu-central-1 eu-west-1 eu-west-2 eu-west-3 sa-east-1)
+# AVAILABLE_REGIONS=(us-east-2 us-east-1 us-west-1 us-west-2 ap-east-1 ap-south-1 ap-northeast-2 ap-southeast-1 ap-southeast-2 ap-northeast-1 ca-central-1 eu-north-1 eu-central-1 eu-west-1 eu-west-2 eu-west-3 sa-east-1 af-south-1 eu-south-1 me-south-1)
```
nhinsch, line 14: What's this comment for?

scripts/add_new_region.sh
```diff
@@ -0,0 +1,62 @@
...
+    # get latest version
+    last_layer_version=$(get_max_version $layer_name $OLD_REGION)
+
+    # run for each version of layer
+    for i in $(seq 1 $last_layer_version); do 
+        layer_path=$layer_name"_"$i.zip
+        aws_version_key="${NODE_VERSIONS_FOR_AWS_CLI[$j]}"
+
+        # download layer versions 
+        URL=$(AWS_REGION=$OLD_REGION aws lambda get-layer-version --layer-name $layer_name --version-number $i --query Content.Location --output text)
```
DarcyRaynerDD, line 54: This is smart, although it might take a while to run.

scripts/check_regions.sh
```diff
@@ -0,0 +1,19 @@
...
+        last_layer_arn=$(aws lambda list-layer-versions --layer-name $layer_name --region $region | jq -r ".LayerVersions | .[0] |  .LayerVersionArn")
+        if [ "$last_layer_arn" == "null" ]; then
+             echo "No layer found for $region, $layer_name"
+             FOUND_REGIONS_MISSING_LAYERS=true
+        fi
+    done
+done
+
+if [ $FOUND_REGIONS_MISSING_LAYERS = true ]; then
+    echo "Run ./add_new_region.sh <new_region> to add layers to the missing regions"
```
DarcyRaynerDD, line 18: Maybe return a fail error code here?

scripts/publish_prod.sh
```diff
@@ -37,3 +37,6 @@ git push origin "refs/tags/v$PACKAGE_VERSION"
 echo 'Publishing Lambda Layer'
 ./scripts/build_layers.sh
 ./scripts/publish_layers.sh
+
+echo 'Checking Regions'
+./scripts/check_regions.sh
```
DarcyRaynerDD, line 42: We should run check_regions before publishing, that way we won't accidentally publish a new version 1 of the layer in the new region, which I think would break the add_region script.

scripts/check_regions.sh
```diff
@@ -0,0 +1,19 @@
...
+        if [ "$last_layer_arn" == "null" ]; then
+             echo "No layer found for $region, $layer_name"
+             FOUND_REGIONS_MISSING_LAYERS=true
+        fi
+    done
+done
+
+if [ $FOUND_REGIONS_MISSING_LAYERS = true ]; then
+    echo "Run ./add_new_region.sh <new_region> to add layers to the missing regions" 
+fi
```
nhinsch, line 19: It looks like this prints warnings if we have missing regions -- is that going to be conspicuous enough to get our attention when we publish the layer?

We could put `WARNING:` at the beginning of the message but I wonder if people still might miss it...

nhinsch: I like Darcy's idea of actually failing here, if that is possible

scripts/add_new_region.sh
hghotra: I think you only do this once when you're publishing to a new region so the delay should be okay considering this will ensure all the versions are correct.

scripts/list_layers.sh
hghotra: ✂️

scripts/publish_prod.sh
hghotra: ✅

scripts/check_regions.sh
hghotra: ✅

PR #103, opened 2020-09-29
Disable the `fs` module integration by default in the tracer

Czechh:
### What does this PR do?
- Disables by default the `fs` module integration in `dd-trace`.
- Removes `http-errors` from integration tests as they really don't provide a lot of value and timeout sporadically because they have an external dependency

In order to enable a user can set `DD_TRACE_DISABLED_PLUGINS` to `''` otherwise it will be disabled by default.

Since the user can already disable other integration in the tracer using `DD_TRACE_DISABLED_PLUGINS` we make sure we don't overwrite the env var and only append to it if it exists.

### Motivation
This confuses a lot of customers and the amount of data we log. This will decrease the number of support cases as well.

### Types of Changes

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

### Check all that apply

- [X] This PR impacts documentation, and it has been updated (or a ticket has been logged)
- [X] This PR passes the integration tests (ask a Datadog member to run the tests)


src/handler.ts
```diff
@@ -3,6 +3,14 @@ import { datadog, datadogHandlerEnvVar, lambdaTaskRootEnvVar, getEnvValue } from
 // tslint:disable-next-line:no-var-requires
 const { load } = require("/var/runtime/UserFunction") as any;
 
+if (getEnvValue("DD_FS_INTEGRATION_ENABLED", "false").toLowerCase() === "false") {
+  if (process.env.DD_TRACE_DISABLED_PLUGINS) {
+    process.env.DD_TRACE_DISABLED_PLUGINS = `${process.env.DD_TRACE_DISABLED_PLUGINS},fs`;
```
nhinsch, line 8: It seems that this could result in `fs` being included in the list twice. I.e. if a customer adds it to the disabled list, but also does not set `DD_FS_INTEGRATION_ENABLED`.

It also might be confusing if a customer adds `fs` to dd_trace_disabled_plugins, but then also adds `DD_FS_INTEGRATION_ENABLED`. The result would be that it would not be enabled, even though DD_FS_INTEGRATION_ENABLED is true; the two env vars conflict. That might be pretty confusing.

Maybe a simpler approach would be to set the default value of DD_TRACE_ENABLED_PLUGINS to `fs`? And allow the customer to pass in a different list if desired?

nhinsch: Sorry, what I said doesn't quite make sense, let me rethink.

Czechh: Great point. Going to make the change to default the env var if none is set.

Czechh: It does make sense. What we can do here is just default the env var to `fs` if none is set.

nhinsch: The main thing is that if a customer passes in a list of disabled plugins that includes `fs`, `fs` should be disabled. And conversely, if they pass in a list of disabled plugins that does not include `fs`, `fs` should not be disabled.

README.md
```diff
@@ -104,6 +104,10 @@ Inject Datadog trace id into logs for correlation. Defaults to `true`.
 
 Set to `true` to merge the X-Ray trace and the Datadog trace, when using both the X-Ray and Datadog tracing. Defaults to `false`.
 
+### DD_TRACE_DISABLED_PLUGINS
```
tianchu, line 107: Since this is not directly supported by `datadog-lambda`, Should we simply add a paragraph about the environment variables that are honored by `dd-trace` and a link to that list? https://docs.datadoghq.com/tracing/setup/nodejs/#configuration

Czechh: Good call - removed this and added a paragraph in `Trace` https://github.com/DataDog/datadog-lambda-js/pull/103/commits/d93ca224074798025631ca4df027e3179b7dc3a2

README.md
```diff
@@ -26,6 +26,8 @@ For additional details on trace collection, take a look at [collecting traces fr
 
 For additional details on trace and log connection, check out the [official documentation for Datadog trace client](https://datadoghq.dev/dd-trace-js/).
 
+The `fs` module is disabled by default. If you want to enable it you have to set the environmental variable `DD_TRACE_DISABLED_PLUGINS` to `''` or to a comma separated list of the plugins you want to disable. See the full list of supported plugins [here](https://docs.datadoghq.com/tracing/compatibility_requirements/nodejs/).
```
nhinsch, line 29: environmental variable -> environment variable
comma separated list -> comma-separated list

nhinsch (approved):

PR #107, opened 2020-10-07
add tags to listener

meghna-dd:
### What does this PR do?

Add version tags to trace/listerner.ts rather than enhanced-metrics.

### Motivation

Better debugging experience.

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


Czechh (approved):
Code LGTM - I'm interested in the integration tests + manual verification before merging+releasing

src/trace/listener.ts
```diff
@@ -9,6 +9,8 @@ import { didFunctionColdStart } from "../utils/cold-start";
 import { Source } from "./constants";
 import { patchConsole } from "./patch-console";
 import { SpanContext, TraceOptions, TracerWrapper } from "./tracer-wrapper";
+const datadogLambdaVersion = "3.32.0";
```
DarcyRaynerDD, line 12: You could export these are reuse them in the spec file, so you only have to substitute once. Or put them in the constants.ts file. I'd also leave this as a dummy value, eg 'x.xx.x' so if we accidentally push a build with the wrong version we can catch it.

scripts/build_layers.sh
```diff
@@ -11,6 +11,13 @@ set -e
 LAYER_DIR=".layers"
 LAYER_FILES_PREFIX="datadog_lambda_node"
 NODE_VERSIONS=("10.15" "12.13")
+DATADOG_LAMBDA_VERSION=$(cat package.json | jq -r ".version")
+DD_TRACE_VERSION=$(cat package.json | jq -r ."devDependencies"'."dd-trace"')
+
+sed -i "" -E "s/(datadogLambdaVersion = )\"([0-9]*\.[0-9]*\.[0-9]*)\"/\1\""$DATADOG_LAMBDA_VERSION"\"/" ./src/trace/listener.ts
```
DarcyRaynerDD, line 17: So, we could apply this in build_layers but then running just yarn build, (which we do when deploying the npm package), wouldn't work. You could create a new script, say './update_dist_version.sh', and run that as a post-script to yarn build eg. https://gist.github.com/coryhouse/b7ff5e6d1e41b762c35b, and modify the dist directory instead of touching the src code every time.

scripts/build_layers.sh
```diff
@@ -11,6 +11,13 @@ set -e
 LAYER_DIR=".layers"
 LAYER_FILES_PREFIX="datadog_lambda_node"
 NODE_VERSIONS=("10.15" "12.13")
+DATADOG_LAMBDA_VERSION=$(cat package.json | jq -r ".version")
+DD_TRACE_VERSION=$(cat package.json | jq -r ."devDependencies"'."dd-trace"')
```
DarcyRaynerDD, line 15: The version of dd-trace-js specified in the package.json can be just a suggestion. For instance ^0.20.0, would mean any minor version greater than 20, but not a major version, so 0.25.5 would match, but not 1.21.0. Right now it's locked in the package.json to an explicit version, but we might change that in the future. The place to actually search is the yarn.lock file, which usually has an entry like this: https://github.com/DataDog/datadog-lambda-js/blob/master/yarn.lock#L1341 . I think we can leave it for now though.

DarcyRaynerDD (changes requested):

src/trace/listener.ts
meghna-dd: Done!

scripts/build_layers.sh
meghna-dd: Done!

scripts/build_layers.sh
meghna-dd: Done!

DarcyRaynerDD (approved):
LGTM

PR #109, opened 2020-10-08
Darcy.rayner/statsd support

DarcyRaynerDD:
### What does this PR do?

Adds statsd/extensions support to the lambda


src/metrics/listener.ts
```diff
@@ -50,13 +52,28 @@ export interface MetricsConfig {
...
+  private isAgentRunning?: boolean = undefined;
 
   constructor(private kmsClient: KMSService, private config: MetricsConfig) {
     this.apiKey = this.getAPIKey(config);
   }
 
-  public onStartInvocation(_: any) {
+  public async onStartInvocation(_: any) {
+    if (this.isAgentRunning === undefined) {
+      this.isAgentRunning = await isAgentRunning();
```
Czechh, line 64: So we would make the request for non-extension users as well? Could we benefit from short-circuit this with an env var and avoid the call?

Czechh (approved):
Looks good! The comment is non-blocker and we can re-visit post launch

src/metrics/listener.ts
DarcyRaynerDD: That's a good point, Remy did suggest we add an env var to control this as well. I think we should revisit for all the layers.

PR #112, opened 2020-10-09
Allow for trace context to be extracted from `clientContext`

Czechh:
### What does this PR do?
This PR adds a fallback in the trace context gathering. The fallback will use the `clientContext` object that is passed as a key in the context for the Lambda execution. More info here: https://docs.aws.amazon.com/lambda/latest/dg/nodejs-context.html

### Motivation

Using this will allow us to inject the trace ids from a caller without modifying the Payload directly. A common problem today is having distributed tracing across multiple chained lambda functions, with this approach we'll be able to do that.

### Testing Guidelines

- Units
- Integration

### Types of Changes

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

### Check all that apply

- [X] This PR contains breaking changes that are documented in the description


tianchu:
I thought `clientContext` is only meant for mobile applications? At least the AWS doc suggests that. How do you inject tracing headers to SDK lambda invoke requests so they appear in `clientContext`? 

BTW, the caller Lambda could be written in languages other than Node.js, does AWS SDK for Python, Node.js (and others) all provide you a way to inject headers to `clientContext`?

Czechh:
@tianchu - Modifying the headers of the AWS Request doesn't seem to work in this case because the Lambda function has only access to the env var, the payload and the context. I started looking at the payload and we could definitely go that route, but it modifies what some customers have build directly. In my experience unstructured payloads are common, and no `header` key is normally present so we would need to account for side effects on our customers business logic if we added that to the Lambda payload.

So I started looking into `clientContext` and some docs do mention that it's for mobile apps but, I think it's because it's a big use-case rather than a limitation. It can be set in any AWS SDK language as it's a parameter used to `invoke`.

**The one gotcha is that it only works for `RequestResponse` calls. So Async calls (`Event` invocations) wouldn't have this info.**

To inject this info we would need to modify `dd-trace-js` to inject the trace `ids` when we call `.invoke` so we pass a `clientContext` in the parameters like so:

```js
const getHeaders = () => {
  const span = tracer.scope().active();
  if (span === null) {
    return {};
  }
  const parentID = span.context().toSpanId();
  const traceID = span.context().toTraceId();

  return {
    'x-datadog-trace-id': traceID,
    'x-datadog-parent-id': parentID,
  };
}

const getClientContext = () => {
  const context = {
    Custom: getHeaders(),
  }
  return Buffer.from(JSON.stringify(context)).toString('base64');
};


exports.handler = (event) => {
  const params = {
    FunctionName: "ironmaiden-dev-ironmaiden",
    Payload: JSON.stringify(createPayload()),
    InvocationType: "Event",
    ClientContext: getClientContext(),
  };

  return lambda.invoke(params).promise();
}
```
Then with this code we can extract the trace `ids` as a fallback for `sdk-to-lambda` calls.

Czechh:
Closing this POC in favor of https://github.com/DataDog/datadog-lambda-js/pull/147

PR #124, opened 2020-11-10
Fix bug with closing xray socket too early

DarcyRaynerDD:
### What does this PR do?

Fixes an issue with the x-ray socket being closed too early. This causes metadata to not be submitted to x-ray.

### Motivation

Discovered while testing

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/trace/context.ts
+++ b/src/trace/context.ts
@@ -135,13 +135,11 @@ export function sendXraySubsegment(segment: string) {
     client = createSocket("udp4");
     // Send segment asynchronously to xray daemon
     client.send(message, 0, message.length, port, address, (error, bytes) => {
+      client?.close();
       logDebug(`Xray daemon received metadata payload`, { error, bytes });
     });
   } catch (error) {
     logDebug("Error occurred submitting to xray daemon", { error });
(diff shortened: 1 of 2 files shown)


src/trace/context.ts
```diff
@@ -135,13 +135,11 @@ export function sendXraySubsegment(segment: string) {
     client = createSocket("udp4");
     // Send segment asynchronously to xray daemon
     client.send(message, 0, message.length, port, address, (error, bytes) => {
+      client?.close();
       logDebug(`Xray daemon received metadata payload`, { error, bytes });
     });
   } catch (error) {
     logDebug("Error occurred submitting to xray daemon", { error });
```
Czechh, line 142: Should we close the socket inside the catch as well? Make sense to close it inside the callback in the happy path, but in the case where an exception is thrown I think closing the socket would make sense as well.

DarcyRaynerDD: I've re-added this client.close

Czechh (approved):

PR #125, opened 2020-11-11
Fix tracing so handler errors are tracked

DarcyRaynerDD:
### What does this PR do?

Fixes bug where root span of trace wouldn't be marked as an error when an exception is thrown.

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


src/index.ts
```diff
@@ -112,34 +112,30 @@ export function datadog<TEvent, TResult>(
...
           localResult = (await promHandler(localEvent, localContext)) as TResult | undefined;
         } catch (err) {
           if (finalConfig.enhancedMetrics) {
             incrementErrorsMetric(metricsListener, localContext);
           }
-          localError = err;
-          localDidError = true;
+          throw err;
+        } finally {
+          await metricsListener.onCompleteInvocation();
```
Czechh, line 131: Can we test that this function is called independently of a rejection from `promHandler`. Ideally a test that failed before and now passes to improve confidence in the change.

DarcyRaynerDD: Good point, I'll add another test case

Czechh (approved):
sorry i meant a unit, but this also works 👍

DarcyRaynerDD:
@Czechh Ahh, yeah. I should have mentioned, I think we already have a test case making sure everything get's called during cleanup when there is an error in the handler. This logic change shouldn't alter that behavior. The actual change was to make sure an exception was thrown through the tracer.wrapped function, (before we were returning a struct with an error or the result from the handler). This insures the error is attached to the trace. I added the integration tests, because it's the only way to verify the change in the trace output.

PR #140, opened 2020-12-18
Extract span tags from triggering event

DylanLovesCoffee:
### What does this PR do?

Parses the event invoking the function for tags to be added to the span. Currently only adds the following services to be detected: API Gateway, ALB, SQS, SNS, DynamoDB, Kinesis, Cloudfront, CW Logs, CW Events. API Gateway and ALB trigger events will also add the HTTP url, method, path, referer and status code response if available.

For the case of X-Ray <> Datadog hybrid tracing, we create an additional dummy Datadog subsegment for the trigger tags.

### Motivation

SLS-790 - Display function trigger span tags

### Testing Guidelines

Unit tests

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


src/trace/listener.ts
```diff
@@ -31,12 +32,21 @@ export interface TraceConfig {
...
 
   public get currentTraceHeaders() {
     return this.contextService.currentTraceHeaders;
   }
+  public get currentSpan() {
+    return this.tracerWrapper.currentSpan;
+  }
+  public get hasHTTPTriggerSource() {
+    return (
+      this.triggerTags && this.triggerTags["trigger.event_source"] === ("api-gateway" || "application-load-balancer")
```
tianchu, line 47: You only want to return `true` when trigger source is `api-gateway`?

src/index.ts
```diff
@@ -122,12 +123,20 @@ export function datadog<TEvent, TResult>(
...
+    let localResult: any | undefined;
     let error: any;
     let didThrow = false;
 
     try {
-      result = await traceListener.onWrap((localEvent: TEvent, localContext: Context) => {
-        return promHandler(localEvent, localContext);
+      result = await traceListener.onWrap(async (localEvent: TEvent, localContext: Context) => {
+        try {
+          localResult = await promHandler(localEvent, localContext);
```
tianchu, line 133: This feels like a significant change to me, should this be done within the `promisifiedHandler`?

src/trace/listener.ts
```diff
@@ -107,6 +118,11 @@ export class TraceListener {
       ) {
         options.tags["_dd.parent_source"] = this.contextService.traceSource;
       }
+      if (this.triggerTags) {
+        options.tags = { ...options.tags, ...this.triggerTags };
+        // # Add trigger tags under the dd subsegment's root_span_metadata field
+        addXrayMetadata(xrayBaggageSubsegmentKey, this.triggerTags);
```
tianchu, line 124: Is this the right place to call this function? I feel the surrounding code only handles the DD native tracing mode. Perhaps should be in `onStartInvocation`?

src/index.ts
```diff
@@ -122,12 +123,20 @@ export function datadog<TEvent, TResult>(
...
+        } finally {
+          if (traceListener.hasHTTPTriggerSource) {
+            setHTTPStatusCodeTag(traceListener.currentSpan, localResult);
```
tianchu, line 136: Does this work for X-Ray?

tianchu:
@DarcyRaynerDD Can you help review as well and give @DylanLovesCoffee some feedback, as I'm not super sure about a few key changes. Hope you can share some thoughts.

src/trace/trigger.ts
```diff
@@ -0,0 +1,270 @@
...
+    triggerTags = { ...triggerTags, ...extractHTTPTags(event) };
+  }
+  return triggerTags;
+}
+
+/**
+ * extractHTTPStatusCode extracts a status code from the response if the Lambda was triggered
+ * by API Gateway or ALB
+ */
+export function extractHTTPStatusCodeTag(triggerTags: { [key: string]: string }, result: any) {
```
DylanLovesCoffee, line 254: I wasn't sure how to type guard the `result` parameter here.

src/trace/trigger.ts
```diff
@@ -0,0 +1,270 @@
...
+  if (!isHTTPTriggerEvent(eventSource)) {
+    return;
+  }
+
+  let statusCode = "200";
+  // Return a 502 status if no response is found
+  if (result === undefined) {
+    statusCode = "502";
+  } else if (result.statusCode) {
+    // Use the statusCode if available
```
DarcyRaynerDD, line 266: Using any for the result type is fine for this method, but we should be stricter with the type checking, especially since this method is called during a finally block and should never throw.

```typescript
const statusCode = result?.statusCode; // Will be undefined if statusCode doesn't exist or result isn't an object

if (typeof statusCode === 'number') {
   return statusCode.toString();
} else if (typeof statusCode ==="string") {
   return statusCode;
}
```

src/trace/trigger.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+    },
+    {
+      result: {
+        "function_trigger.event_source": "sqs",
+        "function_trigger.event_source_arn": "arn:aws:sqs:us-east-1:123456789012:MyQueue",
+      },
+      file: "sqs.json",
+    },
+  ];
+  it("returns the correct event source and ARN", () => {
```
DarcyRaynerDD, line 92: These tests are good, but trigger.ts only has 78% coverage for some reason. We should add tests for `extractHTTPStatusCodeTag` and get this number close to 100.

src/trace/trigger.ts
```diff
@@ -0,0 +1,270 @@
...
+export function parseEventSourceARN(source: string | undefined, event: any, context: Context) {
+  const splitFunctionArn = context.invokedFunctionArn.split(":");
+  const region = splitFunctionArn[3];
+  const accountId = splitFunctionArn[4];
+  const awsARN = getAWSPartitionByRegion(region);
+  let eventSourceARN: string | undefined;
+
+  // e.g. arn:aws:s3:::lambda-xyz123-abc890
+  if (source === "s3") {
+    eventSourceARN = extractS3EventARN(event);
```
DarcyRaynerDD, line 138: The event structure may still be malformed, even after we apply our typeguard. We should add a try catch around this method

DarcyRaynerDD (changes requested):

src/trace/trigger.ts
DylanLovesCoffee: TIL about optional chaining! I kept the check for `result === undefined` here since we would still need to return a `502` status when the Lambda doesn't receive a response.

src/trace/constants.ts
```diff
@@ -16,6 +16,7 @@ export const samplingPriorityHeader = "x-datadog-sampling-priority";
 export const xraySubsegmentName = "datadog-metadata";
 export const xraySubsegmentKey = "trace";
 export const xrayBaggageSubsegmentKey = "root_span_metadata";
+export const xrayLambdaFunctionTagsKey = "lambda_function_tags";
```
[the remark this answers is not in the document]
DylanLovesCoffee: This is mainly just to differentiate the behavior in the [X-Ray converter](https://github.com/DataDog/dd-go/pull/30911/files#diff-c951af2715d3c74a480ce6eae583963d1895b90f0dd2742acb4c66b652b2000cR420-R436). Only the first `root_span_metadata` subsegment that's found is added to the root span, while all `lambda_function_tags` subsegments found are added to their respective parent Lambda execution span(s). On top of that, if X-Ray for API Gateway is enabled, the API Gateway spans are considered the root span.

src/trace/trigger.ts
DylanLovesCoffee: I was curious how you knew that the S3 event structure might be malformed versus others? I wasn't sure if I should try wrapping the other similar methods with a `try...catch`

DarcyRaynerDD: I should have clarified, they could all be malformed, I meant a generic try catch around the entire method.

DarcyRaynerDD (approved):

PR #147, opened 2021-01-11
Extract trace context from client context for Lambda call trace propagation

Czechh:
### What does this PR do?

Adds Lambda client context trace extraction to improve distributed tracing in Serverless environments. 

### Testing Guidelines

- Units
- Integration
- Will test also in Sandbox pre-release 
### Types of Changes

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

### Check all that apply


src/trace/context.ts
```diff
@@ -188,6 +194,39 @@ export function readTraceFromSQSEvent(event: SQSEvent): TraceContext | undefined
...
+  }
+  const parentID = traceData[parentIDHeader];
+  if (typeof parentID !== "string") {
+    return;
+  }
+  const sampledHeader = traceData[samplingPriorityHeader];
+  if (typeof sampledHeader !== "string") {
+    return;
+  }
+  const sampleMode = parseInt(sampledHeader, 10);
```
agocs, line 220: Is there any risk of getting a NaN or otherwise unknown sampleMode? Do we handle that somewhere else?

Czechh: Yeah, this risk exists here. We are assuming that since this is produced in the dd-trace library then it's well formed. If this ends up being NaN then the API will reject the trace. I think though it gets tricky if we want to validate everything before sending. The trace ids could also be malformed.

PR #151, opened 2021-01-13
[POC] Add `DD_TRACE_EXTRACTOR` Env var to customize Trace extraction customization

Czechh:
### What does this PR do?

Provides a way to customize how to extract the trace from the `event` or `context`. This will allow us to quickly support new tools and technologies with custom code.

Here is an example of a caller injecting trace data and an extractor pulling this data to connect the traces:
```js
const tracer = require("dd-trace");
const AWS = require("aws-sdk");
const lambda = new AWS.Lambda();

const createClientContext = data => Buffer.from(JSON.stringify(data)).toString('base64')

exports.caller = (event) => {
  const span = tracer.scope().active();
  const _datadog = {}
  tracer.inject(span, 'text_map', _datadog)

  return lambda.invoke({
    FunctionName: 'ironmaiden-dev-receiver',
    Payload: '{}',
    ClientContext: createClientContext({ custom: { _datadog } }),
  }).promise()
    .then(res => {
      return JSON.stringify(res);
    })
    .catch(err => {
      return JSON.stringify(err);
    });
};

const delay = () => new Promise(resolve => setTimeout(resolve, 1000))

exports.receive = async (event, context) => {
  await delay()

  return { message: 'hallowed be thy name'}
}

exports._extractor = (event, context) => {
  const traceData = context.clientContext.custom._datadog;
  const traceID = traceData["x-datadog-trace-id"];
  const parentID = traceData["x-datadog-parent-id"];
  const sampledHeader = traceData["x-datadog-sampling-priority"];
  const sampleMode = parseInt(sampledHeader, 10);

  const result = {
    parentID,
    sampleMode,
    source: 'event',
    traceID,
  };
  return result
}
```

This would lead to a connect trace in Datadog:
image

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


src/index.ts
```diff
@@ -242,6 +243,16 @@ function getConfig(userConfig?: Partial<Config>): Config {
     const result = getEnvValue(mergeXrayTracesEnvVar, "false").toLowerCase();
     config.mergeDatadogXrayTraces = result === "true";
   }
+  if (userConfig === undefined || userConfig.traceExtractor === undefined) {
+    try {
+      const extractorEnv = getEnvValue(traceExtractorEnvVar, "");
+      const taskRootEnv = getEnvValue(lambdaTaskRootEnvVar, "");
```
DarcyRaynerDD, line 249: Could this cause a circular import issues? Say my-handler.js -> datadog-lambda-js -> my-handler.js? Also, we shouldn't call require if either value is unset.

Czechh: Moved it to the handler file! So it can be injected as a config too Please review!

src/handler.ts
```diff
@@ -18,4 +19,11 @@ if (getEnvValue("DD_TRACE_ENABLED", "true").toLowerCase() === "true") {
 
 const taskRootEnv = getEnvValue(lambdaTaskRootEnvVar, "");
 const handlerEnv = getEnvValue(datadogHandlerEnvVar, "");
-export const handler = datadog(load(taskRootEnv, handlerEnv) as any);
+const extractorEnv = getEnvValue(traceExtractorEnvVar, "");
+let traceExtractor;
+
+if (extractorEnv) {
```
DarcyRaynerDD, line 25: We could try catch around this and add a warning.

src/trace/listener.ts
```diff
@@ -11,6 +11,8 @@ import { Source, ddtraceVersion } from "./constants";
 import { patchConsole } from "./patch-console";
 import { SpanContext, TraceOptions, TracerWrapper } from "./tracer-wrapper";
 
+export type TraceExtractor = (event: any, context: Context) => TraceContext | undefined;
```
DarcyRaynerDD, line 14: I don't think we need `| undefined` at the end here. Having it be optional in trace config is enough.

DarcyRaynerDD (approved):
LGTM, approving for when tests are fixed

Issue #162, opened 2021-02-05
API Gateway v1 protocol causes 500 response from API gateway (possible regression in 3.46.0)


jamestelfer:
## Expected Behaviour

Given an API Gateway integration with the v1 protocol in use, a simple Lambda function should return 200 OK on a valid route.

## Actual Behaviour

API gateway returned 500. With no change to any other configuration, changing the API Gateway to use protocol version 2 fixed the issue.

Using the function handler directly (without Datadog), the same function code responds with 200 OK for both protocol versions.

## Steps to Reproduce the Problem

  1. Create a Lambda with the following handler code:
      ```javascript
      module.exports.handler = async event => {
        return {
            "isBase64Encoded": false,
            "statusCode": 200,
            "headers": { "Content-Type": "application/json" },
            "multiValueHeaders": { },
            "body": "{}"
        }
      }
      ```
  1. Configure the Lambda with the v3.46.0 layer, and the appropriate handler and environment config.
  1. Connect the created Lambda to an HTTP API Gateway with the integration configured to use protocol version 1.

## Specifications

  - Datadog Lambda Layer version: 3.46.0
  - Node version: 12

## Workaround

Downgrade the layer to the previous release (v3.44.0)

DylanLovesCoffee:
Hey @jamestelfer thanks for letting us know . We're looking into the issue and we'll keep you posted on what we find.

DylanLovesCoffee:
@jamestelfer thanks again for raising this. It should be fixed now with release v3.48.0. I'll close the issue out but let us know if you run into anything!

PR #163, opened 2021-02-05
Fixes HTTP tag extraction for API Gateway v1 events

DylanLovesCoffee:
### What does this PR do?

Makes `extractHTTPTags` return immediately after extracting HTTP tags from the trigger event. 

### Motivation

Fixes a bug where API Gateway V1 events were being parsed as V2, resulting in an uncaught exception. Reported by #162 

### Testing Guidelines

Added an API Gateway v1 sample request to the unit tests.

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/trace/trigger.ts
+++ b/src/trace/trigger.ts
@@ -205,6 +205,7 @@ function extractHTTPTags(event: APIGatewayEvent | APIGatewayProxyEventV2 | ALBEv
     if (event.headers.Referer) {
       httpTags["http.referer"] = event.headers.Referer;
     }
+    return httpTags;
   }
 
   if (eventType.isAPIGatewayEventV2(event)) {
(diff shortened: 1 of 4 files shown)


src/trace/trigger.ts
```diff
@@ -248,7 +250,11 @@ export function extractTriggerTags(event: any, context: Context) {
   }
```
DarcyRaynerDD, line 250: Above this is "parseEventSource". Just want to double check nothing in there can throw either. Other than that looks good.

DarcyRaynerDD (approved):

src/trace/trigger.ts
DylanLovesCoffee: Yup I double checked that as well to see if anything in `parseEventSource` might call a non-existent property too. Should be all good.

PR #166, opened 2021-02-17
Fix incorrect parent id for the aws.lambda span

tianchu:
### What does this PR do?

The `currentTraceHeaders` method calls `currentTraceContext`, which always return the _**current**_ trace context. This method is good for getting the trace headers for the injection to downstream calls, because it always sets the parent id to the currently active Datadog span or X-Ray segment. However, it shouldn't be used to determine the parent of the `aws.lambda` span, because it doesn't matter if the user creates a custom span or segment. The `aws.lambda` span should always be parented to the _**root**_ trace context, i.e., the parent span should either be from the incoming trace context or the X-Ray root segment.

Perhaps this wasn't an issue before, because `getXraySegment` doesn't return anything unless X-Ray active tracing is enabled, but it now always return a segment? But it doesn't matter anymore, `getXraySegment` shouldn't be called when determining the parent of the `aws.lambda` span anyway.

To fix the issue, `rootTraceHeaders` was added to `TraceContextService` and it always return the trace headers from the _**root**_ trace context. When determining the `aws.lambda` span's parent in `listener.js`, call `rootTraceHeaders` instead of `currentTraceHeaders`.

Other changes included:
- Upgrade `dd-trace` to `0.31.0` version to allow the Datadog trace context propagate through AWS SDK `lambda.invoke()`.
- Additional comments
- Additional debugging logs 

### Motivation

When the incoming Lambda event object or context object contains a Datadog trace context, the automatically generated `aws.lambda` span should be parented to it. However, it currently parents the span to the X-Ray segment instead, even when X-Ray active tracing is not enabled. This results the `aws.lambda` span being an orphan (parent id points to a non-existent span). 

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


integration_tests/snapshots/logs/http-requests_node14_with_plugin.log
```diff
@@ -1,4 +1,3 @@
-XXXX-XX-XX XX:XX:XX.XXX	XXXX-XXXX-XXXX-XXXX-XXXX	INFO	Snapshot test http requests successfully made to URLs: https://ip-ranges.datadoghq.com,https://ip-ranges.datadoghq.eu
```
tianchu, line 1: Comparing with the logs below, I don't think this line should be there. Probably missed from previous integration test updates.

integration_tests/snapshots/logs/async-metrics_node12_with_plugin.log
```diff
@@ -11,8 +11,8 @@ START RequestId: XXXX Version: $LATEST
...
 XXXX-XX-XX XX:XX:XX.XXX	XXXX-XXXX-XXXX-XXXX-XXXX	INFO	[dd.trace_id=XXXX dd.span_id=XXXX] Processed SNS request
+{"traces":[[{"trace_id":"XXXX","span_id":"XXXX","parent_id":"XXXX","name":"aws.lambda","resource":"integration-plugin-dev-async-metrics_node12_with_plugin","error":0,"meta":{"_dd.origin":"lambda","service":"integration-plugin-dev-async-metrics_node12_with_plugin","cold_start":"false","function_arn":"XXXX_node12_with_plugin","function_version":"$LATEST","request_id":"XXXX","resource_names":"integration-plugin-dev-async-metrics_node12_with_plugin","datadog_lambda":"XXXX","dd_trace":"XXXX","_dd.parent_source":"xray","function_trigger.event_source":"sns","function_trigger.event_source_arn":"arn:aws:sns:us-east-2:123456789012:sns-lambda"},"metrics":{"_sample_rate":1,"_sampling_priority_v1":2},"start":XXXX,"duration":XXXX,"service":"aws.lambda","type":"serverless"}]]}
```
tianchu, line 15: Perhaps the new version of tracer logs the traces a little bit later than the previous version, not sure, but probably doesn't hurt.

src/trace/trace-context-service.ts
```diff
@@ -33,41 +33,62 @@ export class TraceContextService {
...
+    return {
+      [traceIDHeader]: traceContext.traceID,
+      [parentIDHeader]: traceContext.parentID,
+      [samplingPriorityHeader]: traceContext.sampleMode.toString(10),
+    };
+  }
+
+  // Get the trace headers from the root trace context.
+  get rootTraceHeaders(): Partial<TraceHeaders> {
+    const rootTraceContext = this.rootTraceContext;
```
nhinsch, line 75: It looks like we set the root trace context to being the incoming Datadog trace headers. However, when hybrid tracing is enabled, we want to use the PARENT ID from X-Ray but the TRACE ID from Datadog. Is that accounted for here?

nhinsch: Here is the relevant logic in Go: https://github.com/DataDog/datadog-lambda-go/blob/main/internal/trace/context.go#L45

tianchu: Yes, the root trace context is derived from X-Ray when there is not incoming Datadog trace headers.

DarcyRaynerDD: I think Nick is right about the desired behaviour. Otherwise the parent on the root span swaps between the incoming datadog span, (when it's available), and the root x-ray segment. This was an existing issue, so we don't have to fix it with this PR, but I think we would change the logic here: `https://github.com/DataDog/datadog-lambda-js/blob/main/src/trace/context.ts#L88`, to set the parentId to the x-ray when merge x-ray traces is enabled.

DarcyRaynerDD (approved):
LGTM, Nick's comment is a potential issue we should fix, but I don't think we have to do it in this PR.

PR #181, opened 2021-04-21
Add the matrix strategy to workflow to make integration tests faster

maxday:
### What does this PR do?

This PR should make the integration testing step faster by using the matrix strategy, it will run tests in parallel.

### Motivation

Increase productivity by reducing waiting time

### Testing Guidelines

Codebase was not change but a refactoring for integration testing has been made. This should not break the integration tests and should be fully backward compatible

### Additional Notes

Current average workflow time for the integration tests step : between 13 and 17 minutes
With this PR, the new time is about 8min, see here : https://github.com/DataDog/datadog-lambda-js/actions/runs/771988973

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/scripts/build_layers.sh
+++ b/scripts/build_layers.sh
@@ -11,6 +11,14 @@ set -e
 LAYER_DIR=".layers"
 LAYER_FILES_PREFIX="datadog_lambda_node"
 
+if [ -z "$NODE_VERSION" ]; then
+    echo "Node version not specified, running for all node versions."
+    NODE_VERSIONS=("10.15" "12.13" "14.15")
+else
+    echo "Node version is specified: $NODE_VERSION"
+    NODE_VERSIONS=($NODE_VERSION)
+fi
+
 function make_path_absolute {
     echo "$(cd "$(dirname "$1")"; pwd)/$(basename "$1")"
 }
@@ -41,8 +49,11 @@ rm -rf $LAYER_DIR
 mkdir $LAYER_DIR
 
 
-echo "Building layer for node${NODE_VERSION}"
-docker_build_zip ${NODE_VERSION} $LAYER_DIR/${LAYER_FILES_PREFIX}${NODE_VERSION}.zip
+for current_node_version in "${NODE_VERSIONS[@]}"
+do
+    echo "Building layer for node${current_node_version}"
+    docker_build_zip ${current_node_version} $LAYER_DIR/${LAYER_FILES_PREFIX}${current_node_version}.zip
+done
 
 echo "Done creating layers:"
 ls $LAYER_DIR | xargs -I _ echo "$LAYER_DIR/_"
--- a/scripts/build_layers.sh
+++ b/scripts/build_layers.sh
@@ -10,7 +10,6 @@ set -e
 
 LAYER_DIR=".layers"
 LAYER_FILES_PREFIX="datadog_lambda_node"
-NODE_VERSIONS=("10.15" "12.13" "14.15")
 
 function make_path_absolute {
     echo "$(cd "$(dirname "$1")"; pwd)/$(basename "$1")"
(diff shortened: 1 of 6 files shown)


scripts/build_layers.sh
```diff
@@ -10,7 +10,6 @@ set -e
 
 LAYER_DIR=".layers"
 LAYER_FILES_PREFIX="datadog_lambda_node"
-NODE_VERSIONS=("10.15" "12.13" "14.15")
```
DarcyRaynerDD, line 13: We run this locally while deploying, so I think we probably want to keep the ability to build all layers in a single command. We could have an env var or arg that limits the script to a single runtime. You can see an example of how we do that in the [./publish_layers.sh](https://github.com/DataDog/datadog-lambda-js/blob/main/scripts/publish_layers.sh#L28) script

maxday: thanks, this should be resolved with 944c85e

nhinsch:
It looks like the integration tests (and other CI jobs in the same file) are not running on this PR -- not sure why that is but we should look into that...

scripts/run_integration_tests.sh
```diff
@@ -65,49 +64,48 @@ function remove_stack() {
 }
 trap remove_stack EXIT
 
-echo "Deploying functions"
+echo "Deploying functions for runtime $SERVERLESS_RUNTIME"
```
nhinsch, line 67: Nit: I would add a check with a helpful error message in case we forget to specify `$SERVERLESS_RUNTIME`

maxday:
> It looks like the integration tests (and other CI jobs in the same file) are not running on this PR -- not sure why that is but we should look into that...

Yes, we may want to modify the "on:" trigger in the build.yml file to **push** + **pull_request** ?

nhinsch:
> > It looks like the integration tests (and other CI jobs in the same file) are not running on this PR -- not sure why that is but we should look into that...
> 
> Yes, we may want to modify the "on:" trigger in the build.yml file to **push** + **pull_request** ?

I don't thinnnnk that's the issue. This has been working fine on previous PRs ([example](https://github.com/DataDog/datadog-lambda-js/pull/179)) as far as I can tell.

nhinsch:
Let's say I want to run these tests locally. My understanding is that I now need to pass in three new environment variables:

`RUNTIME=xxx NODE_VERSION=xxx SERVERLESS_RUNTIME=xxx ...run_integration_tests.sh`

In addition to the existing `API_KEY` and `UPDATE_SNAPSHOTS` environment variables.

The formats for the three environment variables are different, but fundamentally all three hold the same value. For example, if I want to test Node 10.x, I need to pass `RUNTIME=node10 NODE_VERSION=10.5 SERVERLESS_RUNTIME=node10.x`.

I think it would be more user-friendly to just pass `NODE_VERSION=10` and then have the script use that single variable to determine the correct runtime, node_version, and serverless_runtime based on a hardcoded map. What do you think?

scripts/run_integration_tests.sh
```diff
@@ -11,7 +11,6 @@ set -e
 # These values need to be in sync with serverless.yml, where there needs to be a function
```
nhinsch, line 11: I would update the comment at the top of this file to document the new expected environment variable(s)

maxday:
> Let's say I want to run these tests locally. My understanding is that I now need to pass in three new environment variables:
> 
> `RUNTIME=xxx NODE_VERSION=xxx SERVERLESS_RUNTIME=xxx ...run_integration_tests.sh`
> 
> In addition to the existing `API_KEY` and `UPDATE_SNAPSHOTS` environment variables.
> 
> The formats for the three environment variables are different, but fundamentally all three hold the same value. For example, if I want to test Node 10.x, I need to pass `RUNTIME=node10 NODE_VERSION=10.5 SERVERLESS_RUNTIME=node10.x`.
> 
> I think it would be more user-friendly to just pass `NODE_VERSION=10` and then have the script use that single variable to determine the correct runtime, node_version, and serverless_runtime based on a hardcoded map. What do you think?

Great idea ! I'll do it

nhinsch:
Another thought: can we also keep the option to run all of the integration tests? In case I want to run them all locally and not test one runtime at a time.

nhinsch:
> It looks like the integration tests (and other CI jobs in the same file) are not running on this PR -- not sure why that is but we should look into that...

Per our Slack convo this is because this PR is opened on a forked repo -- let's switch over to the main repo at some point so we can see how everything works in practice.

PR #183, opened 2021-04-22
Add the matrix strategy to workflow to make integration tests faster

maxday:
### What does this PR do?

This PR should make the integration testing step faster by using the matrix strategy, it will run tests in parallel.

### Motivation

Increase productivity by reducing waiting time

### Testing Guidelines

Codebase was not change but a refactoring for integration testing has been made. This should not break the integration tests and should be fully backward compatible

### Additional Notes

Current average workflow time for the integration tests step : between 13 and 17 minutes
With this PR, the new time is about 6-7min

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -70,7 +70,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [node10, node12, node14]
+        runtime-param: [node10, node12, node14]
     steps:
       - name: Checkout
         uses: actions/checkout@v2
@@ -101,5 +101,5 @@ jobs:
           DD_API_KEY: ${{ secrets.DD_API_KEY }}
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-          NODE_VERSION: ${{ matrix.node-version }}
+          RUNTIME_PARAM: ${{ matrix.runtime-param }}
         run: ./scripts/run_integration_tests.sh
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -68,7 +68,19 @@ jobs:
 
   integration-test:
     runs-on: ubuntu-latest
-
+    strategy:
+      matrix:
+        include:
+          - node-runtime: node10
+            node-version: 10.15
+            serverless-runtime: nodejs10.x
+          - node-runtime: node12
+            node-version: 12.13
+            serverless-runtime: nodejs12.x
+          - node-runtime: node14
+            node-version: 14.15
+            serverless-runtime: nodejs14.x
+       
     steps:
       - name: Checkout
         uses: actions/checkout@v2
@@ -99,4 +111,7 @@ jobs:
           DD_API_KEY: ${{ secrets.DD_API_KEY }}
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          RUNTIME: ${{ matrix.node-runtime }}
+          NODE_VERSION: ${{ matrix.node-version }}
+          SERVERLESS_RUNTIME: ${{ matrix.serverless-runtime }}
         run: ./scripts/run_integration_tests.sh
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -70,17 +70,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        include:
-          - node-runtime: node10
-            node-version: 10.15
-            serverless-runtime: nodejs10.x
-          - node-runtime: node12
-            node-version: 12.13
-            serverless-runtime: nodejs12.x
-          - node-runtime: node14
-            node-version: 14.15
-            serverless-runtime: nodejs14.x
-       
+        node-version: [node10, node12, node14]
     steps:
       - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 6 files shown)


scripts/build_layers.sh
```diff
@@ -10,7 +10,14 @@ set -e
 
 LAYER_DIR=".layers"
 LAYER_FILES_PREFIX="datadog_lambda_node"
-NODE_VERSIONS=("10.15" "12.13" "14.15")
+
+if [ -z "$NODE_VERSION" ]; then
+    echo "Node version not specified, running for all node versions."
+    NODE_VERSIONS=("10.15" "12.13" "14.15")
```
nhinsch, line 16: Can we just use the major version, i.e. `NODE_VERSION=10`? I think that would be easier to remember than `NODE_VERSION=10.15` and won't change as often.

scripts/run_integration_tests.sh
```diff
@@ -26,6 +25,24 @@ script_utc_start_time=$(date -u +"%Y%m%dT%H%M%S")
 
 mismatch_found=false
 
+# Format :
+# [0]: serverless runtime name
+# [1]: nodeje version
```
nhinsch, line 30: Nit: nodeje -> nodejs

.github/workflows/build.yml
```diff
@@ -99,4 +101,5 @@ jobs:
           DD_API_KEY: ${{ secrets.DD_API_KEY }}
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          RUNTIME_PARAM: ${{ matrix.runtime-param }}
```
nhinsch, line 104: Likewise here I think it would be nicer to just pass in a number like `10` -- it's always going to be node, so why type it every time? 🙂

scripts/run_integration_tests.sh
```diff
@@ -55,25 +72,40 @@ input_event_files=$(ls ./input_events)
...
-# Generate a random 8-character ID to avoid collisions with other runs
-run_id=$(xxd -l 4 -c 4 -p < /dev/random)
 
-# Always remove the stack before exiting, no matter what
+
+# Always remove the stacks before exiting, no matter what
 function remove_stack() {
-    echo "Removing functions"
-    serverless remove --stage $run_id
+    for runtime in "${RUNTIMES[@]}"; do
```
nhinsch, line 79: I think it's a bit odd to say `runtime in "${RUNTIMES[@]}"` here because each item in that list contains information other than the runtime (like the run ID). In other words, I think it's odd to say `run_id=$runtime[2]`. Maybe there is a better possible name for this, like `parameters_set in ${PARAMETERS_SETS[@]}`. Then we could say `run_id=$parameters_set[2]`.

scripts/run_integration_tests.sh
```diff
@@ -119,13 +152,16 @@ set +e # Don't exit this script if there is a diff or the logs endpoint fails
...
+        function_name="${handler_name}_node"
+        function_snapshot_path="./snapshots/logs/${handler_name}_${runtime}.log"
+        serverless_runtime=$runtime[0]
+        nodejs_version=$runtime[1]
+        run_id=$runtime[2]
         # Fetch logs with serverless cli, retrying to avoid AWS account-wide rate limit error
         retry_counter=0
         while [ $retry_counter -lt 10 ]; do
-            raw_logs=$(serverless logs --stage $run_id -f $function_name --startTime $script_utc_start_time)
+            raw_logs=$(NODE_VERSION=${!nodejs_version} RUNTIME=$runtime SERVERLESS_RUNTIME=${!serverless_runtime} \
```
nhinsch, line 163: Why do we need to know the node version, runtime, etc. to fetch the logs?

nhinsch:
Looks like this cuts the overall time to run the tests to about 5 minutes. Nice! 👏

scripts/build_layers.sh
maxday: actually those versions are useful to get the right alpine tag to build the docker image 
see : ```docker build -t datadog-lambda-layer-node:$1 . --no-cache --build-arg image=node:$1-alpine```
So I think we can't get rid of them ?

nhinsch: My concern is that there are some cases where we want to manually run this script. And it's going to be annoying to have to look up what the minor version number is every time! 😂 

Maybe in the "invalid version" warning, we could have the error message tell you the possible correct version options.

Another option would be to check the argument and convert to the correct alpine tag:

10 -> 10.5
12 -> 12.13
14 -> 14.15
other -> invalid

nhinsch: I like the second option better but the first option would be OK too

scripts/build_layers.sh
```diff
@@ -10,7 +10,14 @@ set -e
...
+else
+    echo "Node version is specified: $NODE_VERSION"
+    NODE_VERSIONS=($NODE_VERSION)
+fi
```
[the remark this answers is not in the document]
maxday: ✅

scripts/run_integration_tests.sh
maxday: ✅

.github/workflows/build.yml
maxday: ✅

scripts/run_integration_tests.sh
maxday: ✅

scripts/run_integration_tests.sh
maxday: we don't :/ but my understanding is that serverless.yaml is fully evaluated at each serverless cmd invocation.
And since this file is now parametrized with some envs, I have noticed that we have to pass them otherwise we end with ```<   A valid environment variable to satisfy the declaration 'env:NODE_VERSION' could not be found.```
```<   A valid environment variable to satisfy the declaration 'env:SERVERLESS_RUNTIME' could not be found.```

nhinsch: That makes sense!

nhinsch (approved):
Amazing! 💯

PR #185, opened 2021-05-13
Fix bug where flushToLogs and the extension would cause timeouts

DarcyRaynerDD:
### What does this PR do?
Identified issue where extension isn't flushing when 'DD_FLUSH_TO_LOGS' is set to true. This modifies the flow to guarantee the extension is always flushed, and that DD_FLUSH_TO_LOGS is ignored when used with the extension.

### Testing Guidelines

Added a unit test, and tested manually.

### Types of Changes

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/metrics/listener.spec.ts
+++ b/src/metrics/listener.spec.ts
@@ -1,9 +1,13 @@
 import { AWSError, KMS, Request } from "aws-sdk";
 import nock from "nock";
+import mock from "mock-fs";
 
 import { LogLevel, setLogLevel } from "../utils";
+import { AGENT_URL } from "./extension";
 
 import { MetricsListener } from "./listener";
+import StatsDClient from "hot-shots";
+jest.mock('hot-shots');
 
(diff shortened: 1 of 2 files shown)


src/metrics/listener.spec.ts
```diff
@@ -96,6 +100,40 @@ describe("MetricsListener", () => {
 
     expect(spy).toHaveBeenCalledWith(`{"e":1487076708,"m":"my-metric","t":["tag:a","tag:b"],"v":10}\n`);
   });
+  it("sends metrics to statsD when extension is enabled", async () => {
```
nhinsch, line 103: I would reword this so that it is clear that this should happen EVEN WHEN log forwarding is enabled. I.e. the behavior we need to test for is that log forwarding does not happen when the extension is enabled.

nhinsch (approved):

PR #192, opened 2021-06-16
Exclude Lambda Extension spans when using tracer

nhinsch:
### What does this PR do?

Configures the tracer's HTTP plugin to exclude spans representing HTTP calls from the Lambda Library to the Lambda Extension (by checking the URL for the port `:8124`).

Note that this will only work if the customer is using the provided handler function and the handler redirection feature, because the handler is where the library initializes the tracer. If the customer instead initializes the tracer in their own code, they would need to apply this configuration there as well.

Bonus: fix the `publish_sandbox` script and disable X-Ray in the integration tests.

### Motivation

By default, the tracer will create a span when the Lambda Library calls the `/flush` endpoint. Because the `/flush` HTTP call from the Library to the Extension occurs outside of the function invocation, it is created as an entirely separate trace from the function invocation. These extraneous traces are potentially confusing and unhelpful.

### Testing Guidelines

I tested manually. Integration tests for the Extension will also help test this behavior.

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


integration_tests/serverless.yml
```diff
@@ -4,9 +4,6 @@ service: integration-tests-js
 provider:
   name: aws
   region: sa-east-1
-  tracing:
```
nhinsch, line 7: I disabled X-Ray tracing in the integration tests. Due to a bug on the AWS side, the `XRAY` log lines are now improperly formatted.

src/handler.ts
```diff
@@ -15,12 +15,19 @@ if (getEnvValue("DD_TRACE_ENABLED", "true").toLowerCase() === "true") {
...
     tags: {
       "_dd.origin": "lambda",
     },
   });
   logDebug("automatically initialized dd-trace");
+
+  // Configure the tracer to ignore HTTP calls made from the Lambda Library to the Extension
+  tracer.use("http", {
+    // blacklist has been deprecated in favor of blocklist, but we still use it here for
+    // compatibility with older versions of the tracer
```
hghotra, line 28: 👍🏼

integration_tests/serverless.yml
DarcyRaynerDD: Is this the same issue we were seeing with Ruby?

nhinsch: Sounds like it

PR #193, opened 2021-06-18
Remove build dependency on `jq` and `apk`

nhinsch:
### What does this PR do?

Previously, we used `jq` during the build process. This build dependency required installation using `apk`.

### Motivation

I have been having a lot of trouble getting `apk` working in my Docker container locally, likely due to some Docker networking issue. Some cursory Googling reveals that this is a common issue. I noticed that `jq` and therefore `apk` were not needed anyway so I decided to remove the dependency.

### Testing Guidelines

The layers still pass the integration tests.

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


nhinsch:
I realized there is an issue here -- this would install the latest version of dd-trace, which is not necessarily the version specified in the `package.json` as a dev dependency.

nhinsch:
I reopened this PR after implementing a solution that will preserve the old behavior while avoiding use of `jq`.

scripts/move_ddtrace_dependency.js
```diff
@@ -0,0 +1,10 @@
+// Moves the dd-trace dependency from devDependencies to dependencies within package.json.
```
maxday, line 1: 💯

maxday (approved):
LGTM 🚀

PR #199, opened 2021-07-06
Modify KMS import to speed up coldstarts

maxday:
### What does this PR do?

- import a more specific (lighter) package
- move import logic to decrypt (this way if a lambda does not use KMS, we skip the KMS loading time as well
- unit test kms-service

### Motivation

Performance

### Testing Guidelines

Unit tests

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/metrics/kms-service.ts
+++ b/src/metrics/kms-service.ts
@@ -3,7 +3,6 @@
 // tries to decrypt a value.
 
 export class KMSService {
-
   public async decrypt(value: string): Promise<string> {
     try {
       const kmsType = require("aws-sdk/clients/kms");
@@ -16,7 +15,7 @@ export class KMSService {
       }
       return result.Plaintext.toString("ascii");
     } catch (err) {
-      throw Error('Couldn\'t decrypt value');
+      throw Error("Couldn't decrypt value");
     }
   }
 }
--- a/src/metrics/kms-service.ts
+++ b/src/metrics/kms-service.ts
@@ -1,36 +1,21 @@
-// This only imports the type of KMS, not the class. As we don't do an instantiation typescript will compile this away.
-import { KMS } from "aws-sdk";
-
-import { logError } from "../utils";
-
 // In order to avoid the layer adding the 40mb aws-sdk to a deployment, (which is always available
 // in the lambda environment anyway), we use require to import the sdk, and return an error if someone
 // tries to decrypt a value.
 export class KMSService {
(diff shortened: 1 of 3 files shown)


src/metrics/kms-service.ts
```diff
@@ -1,36 +1,21 @@
...
-import { logError } from "../utils";
-
 // In order to avoid the layer adding the 40mb aws-sdk to a deployment, (which is always available
 // in the lambda environment anyway), we use require to import the sdk, and return an error if someone
 // tries to decrypt a value.
 export class KMSService {
-  private kms?: KMS;
 
-  constructor() {
+  public async decrypt(value: string): Promise<string> {
```
DarcyRaynerDD, line 6: If somebody installs the extension and uses datadog kms api key variable, decrypt will still get called even though the library doesn't end up using the key. Might be another potential optimization for clients using KMS.

src/metrics/kms-service.ts
```diff
@@ -1,36 +1,21 @@
...
     } catch (err) {
-      logError("optional dependency aws-sdk not installed. KMS key decryption will not work", err);
-    }
-  }
-
-  public async decrypt(value: string): Promise<string> {
-    if (this.kms === undefined) {
-      // This error should only occur if the user is running this code outside of the lambda environment,
-      // (say a local development environment).
       throw Error("optional dependency aws-sdk not installed. KMS key decryption will not work");
```
DarcyRaynerDD, line 18: I think we should log original error before throwing again.

DarcyRaynerDD (approved):
LGTM, great find

src/metrics/kms-service.ts
maxday: absolutely ! maybe in an other PR ?

src/metrics/kms-service.ts
maxday: ✅

src/metrics/kms-service.ts
DarcyRaynerDD: Sure, I don't think it has to be done right now.

DarcyRaynerDD (approved):
LGTM, nice tests

src/metrics/kms-service.ts
maxday: OK I'll create a ticket for that

PR #202, opened 2021-07-14
Support decrypting API keys that were encrypted without an encryption context

nhinsch:
### What does this PR do?

Previously, we assumed that the `DD_KMS_API_KEY` was encrypted using the AWS console. When the API key is encrypted using the console, the console uses the Lambda function's name as an "encryption context". We need to provide this same encryption context in the decrypt call to successfully decrypt the API key. However, if the customer encrypts the API key using the AWS CLI, it will not have this encryption context and the decrypt call will fail.

With this PR, we will first try decrypting the API key without an encryption context, and then attempt to decrypt it with the decryption context.

### Motivation

Encrypting the API key using the AWS console does not work well with our deployment tools, because we expect the customer to provide a single encrypted API key that will work across multiple functions. Because the AWS console uses the Lambda function's name as the encryption context, each encrypted API key will only work with that individual Lambda function.

### Testing Guidelines

Unit test coverage.

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/scripts/publish_sandbox.sh
+++ b/scripts/publish_sandbox.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Usage: VERSION=5 aws-vault exec sandbox-acccount-admin -- publish_layers.sh
+# Usage: VERSION=5 aws-vault exec sandbox-acccount-admin -- publish_sandbox.sh
 set -e
 
 if [ -z "$VERSION" ]; then
@@ -9,6 +9,6 @@ if [ -z "$VERSION" ]; then
     exit 1
 fi
 
-./scripts/build_layers.sh
-./scripts/sign_layers.sh sandbox
+# ./scripts/build_layers.sh
+# ./scripts/sign_layers.sh sandbox
 VERSION=$VERSION REGIONS=sa-east-1 ./scripts/publish_layers.sh
--- a/src/metrics/kms-service.spec.ts
+++ b/src/metrics/kms-service.spec.ts
@@ -3,6 +3,7 @@ import nock from "nock";
 
 describe("KMSService", () => {
   const ENCRYPTED_KEY = "BQICAHj0djbIQaGrIfSD2gstvRF3h8YGMeEvO5rRHNiuWwSeegEFl57KxNejRn"; // random fake key
+  const KEY_ID = "arn:aws:kms:us-east-1:111122223333:key/1234abcd-12ab-34cd-56ef-1234567890ab";
   const EXPECTED_RESULT = "myDecryptedKey";
   const EXPECTED_ERROR_MESSAGE = "Couldn't decrypt ciphertext";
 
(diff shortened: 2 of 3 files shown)


maxday:
Is there the same issue with the extension?
Having two calls to KMS will slow down the coldstart. Regarding our effort to speed it up, would it be possible to require that KMS usage will imply the function name in the encryption content? 
I see that it is possible to pass an encryption context using the cli with `--encryption-context` (https://docs.aws.amazon.com/cli/latest/reference/kms/encrypt.html).

Should we support only this method so we need only to do one single call? (and maybe bump the major version since it's a breaking change). We also could explain in the doc that this is way more secure, since if someone has access to the encrypted key, it won't be able to copy/paste it to a new function.
What do you think?

maxday (commented):
Is there the same issue with the extension?
Having two calls to KMS will slow down the coldstart. Regarding our effort to speed it up, would it be possible to require that KMS usage will imply the function name in the encryption content?
I see that it is possible to pass an encryption context using the cli with --encryption-context (https://docs.aws.amazon.com/cli/latest/reference/kms/encrypt.html).

Should we support only this method so we need only to do one single call? (and maybe bump the major version since it's a breaking change). We also could explain in the doc that this is way more secure, since if someone has access to the encrypted key, it won't be able to copy/paste it to a new function.
What do you think?

nhinsch:
> Is there the same issue with the extension?

The extension has already been updated to support decryption with or without the encryption context. See [this PR](https://github.com/DataDog/datadog-agent/pull/8328/files). This was released in version 9. 

There is a difference between the extension implementation and the implementation in this PR. The extension tries to decrypt WITH an encryption context first (falling back to WITHOUT encryption context). In this JS library PR, the order is reversed: we try first WITHOUT an encryption context and fall back to WITH an encryption context. That is because encryption WITHOUT the encryption context will be the new preferred method (for reasons I will explain below). I am going to update the extension and other libraries to reverse the order to match this new preferred order.

> Having two calls to KMS will slow down the coldstart. Regarding our effort to speed it up, would it be possible to require that KMS usage will imply the function name in the encryption content?
> I see that it is possible to pass an encryption context using the cli with --encryption-context (https://docs.aws.amazon.com/cli/latest/reference/kms/encrypt.html).
> 
> Should we support only this method so we need only to do one single call? (and maybe bump the major version since it's a breaking change). We also could explain in the doc that this is way more secure, since if someone has access to the encrypted key, it won't be able to copy/paste it to a new function.
> What do you think?

That was the original plan, but the problem is that if the function name is in the encryption context, you can't re-use the encrypted API key across multiple functions, which means you can't use it with our deployment tools (e.g. Serverless Framework plugin). The customer would also have to re-encrypt their API key every time they deploy a function. So, the new plan is to prefer that the customer use an API key encrypted without an encryption context.

If the customer does encrypt the API key without an encryption context, then there will only be one call to KMS.

With regard to security, we do in fact want the customer to be able to re-use the same encrypted key across functions. However, the customer won't simply be able to use the copy/pasted API key anywhere they want. Because the key is encrypted, any resource that wants to use the key will need IAM permissions to decrypt using the customer's chosen customer master key (CMK). This is why encrypting the key using KMS makes it more secure -- it can no longer be freely used by copy/pasting.

maxday:
Thanks for the clarification. Regarding `which means you can't use it with our deployment tools` could we handle this logic in our tools? It should be doable to generate a new KMS key for each function based on their name?

nhinsch:
That could be possible, but we would need the user who is deploying the functions to have the necessary IAM `kms:encrypt` permissions to encrypt the key for each function. Also, we would need to update all of our deployment tools to support generating encrypted keys, and allow the customer to specify which CMK to use for encryption.

maxday (approved):

PR #218, opened 2021-08-17
Properly exclude Lambda Extension spans 

nhinsch:
### What does this PR do?

Switches from the deprecated `blacklist` option in dd-trace-js to the newer `blocklist` when excluding HTTP calls made from dd-lambda-js to the Lambda Extension. The deprecated option is not working properly.

I also updated the blocklist regex from `:8124` to `8124\/lambda`. Both of these regexes work fine for excluding the span, but the new regex will allow other requests made on port 8124 to be traced.

### Motivation

In certain situations, the tracer records a span representing the HTTP call that dd-lambda-js makes to the Lambda Extension at the end of an invocation. It is still unclear to me what exact situations this happens in, but it seems to be caused by differences in the exact order in which modules are imported.

Either way, we want to exclude these spans in all cases. The option we are currently using in dd-trace-js is not working properly, so switching to the newer option will solve the issue for us. I have also reported the issue with the deprecated option to the dd-trace-js team.

### Testing Guidelines

I manually tested 16 different combinations of changes to identify and confirm the cause of the issue and its solution.

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/handler.ts
+++ b/src/handler.ts
@@ -24,7 +24,7 @@ if (getEnvValue("DD_TRACE_ENABLED", "true").toLowerCase() === "true") {
 
   // Configure the tracer to ignore HTTP calls made from the Lambda Library to the Extension
   tracer.use("http", {
-    blocklist: /8124\/lambda/,
+    blocklist: /:8124\/lambda/,
   });
 }
 
--- a/src/handler.ts
+++ b/src/handler.ts
@@ -24,9 +24,7 @@ if (getEnvValue("DD_TRACE_ENABLED", "true").toLowerCase() === "true") {
 
   // Configure the tracer to ignore HTTP calls made from the Lambda Library to the Extension
   tracer.use("http", {
-    // blacklist has been deprecated in favor of blocklist, but we still use it here for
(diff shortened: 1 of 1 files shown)


src/handler.ts
```diff
@@ -24,9 +24,7 @@ if (getEnvValue("DD_TRACE_ENABLED", "true").toLowerCase() === "true") {
 
   // Configure the tracer to ignore HTTP calls made from the Lambda Library to the Extension
   tracer.use("http", {
-    // blacklist has been deprecated in favor of blocklist, but we still use it here for
-    // compatibility with older versions of the tracer
-    blacklist: /:8124/,
+    blocklist: /8124\/lambda/,
```
maxday, line 27: do we want to leave the `:` to ensure that a customer who has an url with uuid/lambda (where uuid would finish with 8124) won't have any conflict? It seems unlikely but having the `:` would be more safe? WDYT?

nhinsch: Sure, why not?

maxday (approved):

PR #222, opened 2021-09-10
Optionally capture Lambda request and response payloads

astuyve:
### What does this PR do?

Uses span tags to recursively tag attrs of the request and response object.
- Max depth of 5 (not configurable)
- Max length of 5,000 chars (backend limit for span tags)
- Filters out common auth patterns
- Allows for more specific patterns with `datadog.yaml` obfuscations

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

- [x] This PR introduces new APIs or parameters that are documented and unlikely to change in the foreseeable future

### TODO
- [ ] Documentation update


src/utils/tagObject.ts
```diff
@@ -0,0 +1,34 @@
+const redactableKeys = ["authorization", "x-authorization", "password", "token"];
```
maxday, line 1: Maybe this could be customizable by the customer?

astuyve: Oh good question - we actually already allow the customer to define obfuscation rules via `datadog.yaml` https://docs.datadoghq.com/tracing/setup_overview/configure_data_security/?tab=datadogyaml#replace-rules-for-tag-filtering

This is just a hardcoded list to prevent non-configured payload capture from indexing basic/common secrets. It's not something we want to automatically index. If a user wants to capture these, they'll have to manually set a tag.

maxday: oh I see, so maybe we should make a link/note about this config option when we'll release this new env var?

astuyve: Yup! And it'll be in the blog post :)

src/utils/index.ts
```diff
@@ -3,3 +3,4 @@ export { wrap, promisifiedHandler } from "./handler";
 export { Timer } from "./timer";
 export { logError, logDebug, Logger, setLogLevel, setLogger, LogLevel } from "./log";
 export { get, post } from "./request";
+export { tagObject } from "./tagObject";
```
DarcyRaynerDD, line 6: nit: we typically name files like this `tag-object.ts`

src/utils/tagObject.ts
```diff
@@ -0,0 +1,40 @@
...
+  } else {
+    depth += 1;
+  }
+  if (obj === null) {
+    return currentSpan.setTag(key, obj);
+  }
+  if (typeof obj === "string") {
+    let parsed: string;
+    try {
+      parsed = JSON.parse(obj);
```
DarcyRaynerDD, line 16: If it's a string, does it need to be parsed?

DarcyRaynerDD (approved):
LGTM, left some comments.

src/utils/tagObject.ts
astuyve: Good question. This is needed to deeply tag nested JSON into separate items, otherwise it just tags an object (ala: `setTag('request.body', { foo: { blah: 'ahhh' }})`, which doesn't display nicely in the app.

PR #223, opened 2021-09-10
fix: no dangling promises

astuyve:
### What does this PR do?

We've got a promise inside a synchronous method. It's unclear to me if we actually care about this promise returning. If not, we can ignore the rule and I will revert the code changes.

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -14,10 +14,8 @@ import { LogLevel, setLogLevel } from "./utils";
 
 jest.mock("./metrics/enhanced-metrics");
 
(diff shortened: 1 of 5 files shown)


nhinsch (approved):
Looks good to me!
We should bump the major version since we are making the public methods `async`

astuyve:
@nhinsch Thanks for the review! After discussing, I will revert the breaking change, re-introduce the rule, and ignore it in this specific place (as to avoid a breaking change)

PR #227, opened 2021-09-29
Add metric to report integration tests status

maxday:
### What does this PR do?

Send a metric at the end of integration tests
- with value of 0 if success
- with value of 1 otherwise

This is done by inspecting the status of the integration tests step :

Screen Shot 2021-09-29 at 2 04 03 PM

Bonus : fix a flaky test (ip for lookup is not consistent)

Screen Shot 2021-09-29 at 1 47 37 PM

### Motivation

Detect (via monitor + alert) when daily runs of integration tests are failing

### Testing Guidelines

Make the test fail
- you should see the alert
Fix it
- you should see the recover

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


.github/workflows/build.yml
```diff
@@ -118,3 +118,20 @@ jobs:
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           RUNTIME_PARAM: ${{ matrix.runtime-param }}
         run: ./scripts/run_integration_tests.sh
+
+      - name: Send success metric
+        env:
+          DD_API_KEY_SERVERLESS: ${{ secrets.DD_API_KEY_SERVERLESS }}
```
maxday, line 124: a new secret was added not to overwrite the existing DD_API_KEY secret but maybe we should change the existing one to point to the serverless Org and use only one DD_API_KEY?

agocs: Do you know the whole DD_API_KEY? I think we could look it up in postgres

.github/workflows/build.yml
```diff
@@ -118,3 +118,20 @@ jobs:
...
+        run: ./scripts/send_status_metric.sh 0 $DD_API_KEY_SERVERLESS
+
+  run-if-failed:
```
nhinsch, line 127: I would name this "integration-test-failure" so we know that it runs when the integration tests failed, not the other tests/linter/etc.

Alternatively, can we make this a step inside the integration tests job instead of a separate job? This doc makes me think it might be possible: https://docs.github.com/en/actions/learn-github-actions/contexts#job-status-check-functions

scripts/run_integration_tests.sh
```diff
@@ -207,7 +207,9 @@ for handler_name in "${LAMBDA_HANDLERS[@]}"; do
                 # Remove metrics and metas in logged traces (their order is inconsistent)
                 perl -p -e 's/"(meta|metrics)":{(.*?)}/"\1":{"XXXX": "XXXX"}/g' |
                 # Normalize enhanced metric datadog_lambda tag
-                perl -p -e "s/(datadog_lambda:v)[0-9\.]+/\1X.X.X/g"
+                perl -p -e "s/(datadog_lambda:v)[0-9\.]+/\1X.X.X/g" |
+                # Normalize lookup ressource
```
nhinsch, line 211: ressource -> resource

nhinsch:
This looks good, but I thought we were going to add this metric for the extension integration tests, not the library integration tests. Since theses tests run on every PR, is there a risk that we wouldn't notice a failure?

.github/workflows/build.yml
maxday: fyi, we are able to confirm that this linked to an incorrect org, I'll replace the secret and use just one

.github/workflows/build.yml
nhinsch: Specifically, I am looking at `steps.<step id>.outcome` and `continue-on-error`

.github/workflows/build.yml
nhinsch: Yeah, I would have guessed that the API key in here would already be for the Serverless org 🤔 I definitely don't think we need two different API keys in the tests.

.github/workflows/build.yml
maxday: Renamed! I think I preferred a separated step as it shows up nicely in the workflow diagram (screenshot in the description)

maxday: > This looks good, but I thought we were going to add this metric for the extension integration tests, not the library integration tests. Since these tests already run on every PR, is there a risk that we wouldn't notice a failure?

Yes I think so, since those tests are also run every day, we might have flaky tests which may have passed during the PR but will fail the day after. What do you think?

nhinsch:
Is this going to run on every commit/PR? I think we only want to send the metric if the integration tests are failing on the master branch.

maxday:
> This looks good, but I thought we were going to add this metric for the extension integration tests, not the library integration tests. Since these tests already run on every PR, is there a risk that we wouldn't notice a failure?

Yes I think so, since those tests are also run every day, we might have flaky tests which may have passed during the PR but will fail the day after. What do you think?

maxday:
> Is this going to run on every commit/PR? I think we only want to send the metric if the integration tests are failing on the master branch.

Good catch! I'll update the hook

.github/workflows/build.yml
```diff
@@ -118,3 +118,20 @@ jobs:
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           RUNTIME_PARAM: ${{ matrix.runtime-param }}
         run: ./scripts/run_integration_tests.sh
+
+      - name: Send success metric
```
nhinsch, line 122: Do we even need a success metric?

Would it be simpler to just send a "failures" metric, and expect the value of the metric to be 0?

That way, the metric can be called "failures" instead of "status". It would work kind of like the other errors metrics that we have, like `lambda.errors`

nhinsch: Or alternatively, if we do want to count successes, maybe it should be a separate `.success` metric?

I think right now the metric is being conceived as a `GAUGE` type where we have a decimal value between 0 (all successes) and 1 (all failures). I think it would be more intuitive to have the metric be a `COUNT` type where we just send a data point each time a failure occurs. (We could optionally have another metric for successes).

https://docs.datadoghq.com/metrics/types/?tab=count#metric-types

Also, do we need to specify the metric type in the API call?

maxday: I found it useful as a condition to recover for an alert in the monitor. Would it be possible to do that with one single metric/value?

nhinsch: Talked about this on Zoom -- I got confused here between the behavior of `GAUGE` and `RATE`. This solution looks good.

nhinsch (approved):

PR #229, opened 2021-10-01
feat: support reading trace context from AppSync event

bigkraig:
### What does this PR do?

This PR adds support for reading the datadog headers from an AppSync event, without requiring users to use custom trace extractors.

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/utils/event-type-guards.ts
+++ b/src/utils/event-type-guards.ts
@@ -1,6 +1,7 @@
 import {
   APIGatewayEvent,
   APIGatewayProxyEventV2,
+  AppSyncResolverEvent,
   ALBEvent,
   CloudWatchLogsEvent,
   ScheduledEvent,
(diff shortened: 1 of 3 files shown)


astuyve:
Hi @bigkraig - thanks so much for the contribution! We'll take a look as soon as we can.

bigkraig:
Thanks @astuyve!

src/utils/event-type-guards.ts
```diff
@@ -58,3 +59,7 @@ export function isSNSEvent(event: any): event is SNSEvent {
 export function isSQSEvent(event: any): event is SQSEvent {
   return Array.isArray(event.Records) && event.Records.length > 0 && event.Records[0].eventSource === "aws:sqs";
 }
+
+export function isAppSyncResolverEvent(event: any): event is AppSyncResolverEvent<any> {
```
maxday, line 63: This check may be a bit permissive? Can we be sure that event.info.variables is used only for AppSync related events? According to the doc there is a `selectionSetGraphQL` field in the event, this might be more tightly coupled with AppSync?

bigkraig: :+1: I can switch to that

maxday:
Thanks for your contribution @bigkraig  👍

PR #233, opened 2021-10-06
feat: allow apiKey config option to be a promise

jbreckel:
* this allows fetching the apiKey during runtime and removes the need to have the key clearly visible in the environment or using a workaround with KMS for multiple accounts and environments

### What does this PR do?

This PR adjusts the type of `config.apiKey` to also include `Promise<string>`.
As the `apiKey` is used in tandem with `apiKeyKms`, changing it to a promise should not introduce any issues.

Allowing the `apiKey` to be a promise allows the user to fetch the `apiKey` from a third party (e.g. SecretsManager without exposing it in a lambda environment.

### Motivation

I want to be able to use an `apiKey` that is stored in SecretsManager by only exposing the arn in the lambda environment.

### Testing Guidelines

I added a unit test and tested it by locally installing the module and deploying it to our test system.

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -120,6 +120,26 @@ describe("datadog", () => {
     expect(nock.isDone()).toBeTruthy();
   });
 
+  it("allow apiKey property to be a promise", async () => {
+    const apiKey = "123456";
+    const apiKeyPromise = Promise.resolve(apiKey);
+
+    nock("https://api.datadoghq.com")
+      .post(`/api/v1/distribution_points?api_key=${apiKey}`, (request: any) => request.series[0].metric === "my-dist")
+      .reply(200, {});
+
+    const wrapped = datadog(
+      async () => {
+        sendDistributionMetric("my-dist", 100, "first-tag", "second-tag");
+        return "";
+      },
+      { forceWrap: true, apiKey: apiKeyPromise },
+    );
+    await wrapped({}, {} as any, () => {});
+
+    expect(nock.isDone()).toBeTruthy();
+  });
+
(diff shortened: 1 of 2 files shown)


jbreckel:
possible workaround:
```
async function getApiKey(): Promise<string> {
  // fetch api key from anywhere
}

export const handler = datadog(handlerFunction), {
  apiKey: getApiKey() as unknown as string,
})
```

tianchu:
@jbreckel really appreciate your contribution! But we have already been making improvement over supporting secrets manager as the source for DD_API_KEY, it's actually supported by the latest version of the Extension already I believe, can you try storing your api key as a plaintext (instead of json structure) in Secretes Manager and then set the ARN as env var `DD_API_KEY_SECRET_ARN` on your lambda?

Also this library normally don't need the Datadog api key, except for a special use case where you only send custom metrics from the lambda, and want to send the metrics over the library synchronously. It's the Extension https://github.com/DataDog/datadog-lambda-extension that needs the API key to send data. So I doubt your change makes a difference or solves your problem.

If you do need this change, I think directly adding the support of `DD_API_KEY_SECRET_ARN` to this library is probably even better, like we have for Python https://github.com/DataDog/datadog-lambda-python#dd_api_key.

Again, thanks for your contribution, please let me know what do you think.

tianchu:
@jbreckel could also be helpful! Also do you mind adding a 👍 or leave a comment on that issue to indicate customer interest? We rely on those signals for prioritization. Thank you!

PR #282, opened 2022-04-19
ESM modules support

DarcyRaynerDD:
### What does this PR do?

Adds support for ESM modules and top level async await support with the lambda layer.

JS files with a '.mjs' extension, or with a package.json containing 'type':'module' field, will be imported as [ESM modules.](https://aws.amazon.com/blogs/compute/using-node-js-es-modules-and-top-level-await-in-aws-lambda/). 

ESM modules look like this:
```js

import {makePromise} from 'my-lib.js';

let prom = makePromise();
await prom(); // Top level await, this module won't finish importing until this is complete.

export async function handler() {
   return {
      status: 200,
      message: "Hello"
   }
}
```
VS CommonJS 

```js
const {makePromise} = require('my-lib');
exports.handler = async function {
   return {
      status: 200,
      message: "Hello"
   }
};
```
### Motivation

This is a supported feature of Lambda, that we are catching up to.

### Testing Guidelines

I've added an integration test for the .mjs use case. II have tested manually for ARM for now.

### Additional Notes

We upgraded the handler wrapper script to use a dynamic imports for mjs modules, (node will throw an error if you try to use require).

Dynamic imports look like this:
```js
const my_lib_req = require('my_lib_commonjs'); // Returns the exported object
const my_lib_prom = import('my_lib.mjs'); // Returns a promise
const my_lib_req = require('my_lib.mjs"); // Throws an error
````
We need a way to depromisify the handler library. Usually this is done be turning the calling function into an async function, and awaiting the result. However, this will be reflected all the way up the call hierachy, until we get to our root function, where we still have a promise. 

The solution is to use the library [deasync](https://www.npmjs.com/package/deasync).  This is a native library, (written in C), that needs to be compiled to use each CPU Architecture, and hooks more directly into the Node runtime. Unfortunately, deasync doesn't come with a prebuilt version of the library for linux-arm64, so we added a script to our build process that will create this. We also strip out other pre-built versions of deasync, since they would add > 6mb to the layer size.

### Types of Changes

- [ ] Bug fix
- [X] New feature
- [X] Breaking change

### Check all that apply


DarcyRaynerDD:
Closing this, we are going to use another approach that doesn't involve using deasync

PR #285, opened 2022-05-16
Node 16 support

DarcyRaynerDD:
### What does this PR do?

Adds support for Node16, with ESM modules and top level async await support with the lambda layer.

JS files with a '.mjs' extension, or with a package.json containing 'type':'module' field, will be imported as [ESM modules.](https://aws.amazon.com/blogs/compute/using-node-js-es-modules-and-top-level-await-in-aws-lambda/). 

ESM modules look like this:
```js

import {makePromise} from 'my-lib.js';

let prom = makePromise();
await prom(); // Top level await, this module won't finish importing until this is complete.

export async function handler() {
   return {
      status: 200,
      message: "Hello"
   }
}
```
VS CommonJS 

```js
const {makePromise} = require('my-lib');
exports.handler = async function {
   return {
      status: 200,
      message: "Hello"
   }
};
```
### Motivation

This is a supported feature of Lambda, that we are catching up to.

### Testing Guidelines

I've added an integration test for the .mjs use case. I have tested manually for ARM for now.

### Additional Notes

We change the handler file to be a ES module, so it can import user ES modules. (You can't require an ES module from a CommonJS module).  We split behavior with the node 12 runtime, and use a commonjs version of the handler file that can't import ES modules. This break backwards compatibility. Customer's using the NPM module with v12 of the lambda and the redirection flow, will need to rename their handler to `/opt/nodejs/node_modules/datadog-lambda-js/handler.cjs.handler`. Customer's using the layer won't need to make any changes.

### Types of Changes

- [ ] Bug fix
- [X] New feature
- [x] Breaking change

### Check all that apply

- [X] This PR contains breaking changes that are documented in the description


integration_tests/snapshots/return_values/esm_node12_api-gateway-get.json
```diff
@@ -0,0 +1 @@
+Invocation failed
```
astuyve, line 1: This should be successful and contain a return value, right?

PR #293, opened 2022-06-28
Send errors metrics for 5xx response for Node.js

kimi-p:
https://datadoghq.atlassian.net/browse/SLS-1775

### What does this PR do?
This PR sets Error column on the invocations table and shows error on the Lambda span (not on API Gateway span because then the error counts wouldn't show up on lambdas of the summary view).

image

image

### Motivation
Requests of 5xx status code should be treated as errors.

### Testing Guidelines
Unittest and sandbox testing.

### Additional Notes
Corresponding Frontend change: https://github.com/DataDog/web-ui/pull/60488/files

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

- [] This PR's changes are covered by the automated tests


src/index.ts
```diff
@@ -136,7 +136,7 @@ export function datadog<TEvent, TResult>(
         try {
           localResult = await promHandler(localEvent, localContext);
         } finally {
-          traceListener.onEndingInvocation(localEvent, localResult, finalConfig.captureLambdaPayload);
+          traceListener.onEndingInvocation(localEvent, localResult, metricsListener, localContext, finalConfig.captureLambdaPayload);
```
kimi-p, line 139: Not sure if doing the logic in tracerListener is the best idea, but duplicating the logic from tracerListener to here is probably worse. Any suggestions are appreciated.

maxday: What about : 
- making `onEndingInvocation` returning a boolean (true for success, false otherwise)
- call `incrementErrorsMetric(metricsListener, context);` in index (depending on the bool result)
This way you don't add coupling between trace and metric listeners?
WDYT?

kimi-p: Sounds great! This approach is much cleaner.

astuyve: Alternatively you could add a new hook for the metrics listener called `onEndingInvocation` and pass the local result/context there. Either way makes sense I think.

kimi-p:
retest

kimi-p:
Adding `get`, `delete`, and `post` access to the github action user in the sandbox account.
Policy name: `Serverless-Integration-Test-Github-Action-User-Post-API-Gateway`

integration_tests/snapshots/return_values/status-code-500s_node12_api-gateway-get.json
```diff
@@ -0,0 +1,4 @@
+{
+    "statusCode": 501,
```
maxday, line 2: filenames are containing 500 but it seems that it's returning 501?

src/index.spec.ts
```diff
@@ -77,6 +77,7 @@ describe("datadog", () => {
     expect(parentId).toEqual("9101112");
     expect(sampled).toEqual("2");
   });
+
```
maxday, line 80: could you revert this change?

src/trace/listener.ts
```diff
@@ -139,8 +140,14 @@ export class TraceListener {
       }
       if (this.inferredSpan) {
         this.inferredSpan.setTag("http.status_code", statusCode);
+
+        if (statusCode?.length === 3 && statusCode?.startsWith("5")) {
+          this.wrappedCurrentSpan.setTag("error", "5xx Server Error");
```
maxday, line 145: is it normal that this tag does not show up in the snapshots?

integration_tests/status-code-500s.js
```diff
@@ -0,0 +1,13 @@
+const tracer = require("dd-trace").init();
+const { datadog } = require("datadog-lambda-js");
+
+async function handle(event, context) {
+  const span = tracer.scope().active();
+  span.setTag("fake_tag", "fake_value")
```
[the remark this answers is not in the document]
kimi-p: I was testing if the tag would show up or not. Removed.

src/trace/listener.ts
kimi-p: Good question. I thought about this too. I think `error` tag is treated differently. Asked in slack, if no one knows other reasons, I think special treatment is probably the reason.

kimi-p: Updated the value of this tag to `1`. The theory is that this `error` tag is transformed to ` "error": 1` instead of `meta` field like other normal tags.
Tested the change in sandbox and works as expected (and the same as before).
image

maxday (approved):
LGTM! 🎉

PR #297, opened 2022-08-02
Change error logs to debug for trigger type parsing

IvanTopolcic:
### What does this PR do?

These logs are not technically errors, as they're expected flow in some cases. For example, event bridge triggers do not necessarily need to be ARNs.

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/trace/trigger.ts
+++ b/src/trace/trigger.ts
@@ -14,7 +14,7 @@ import {
   EventBridgeEvent,
 } from "aws-lambda";
 import * as eventType from "../utils/event-type-guards";
-import { logError } from "../utils";
+import { logDebug } from "../utils";
 import { gunzipSync } from "zlib";
 
 type LambdaURLEvent = {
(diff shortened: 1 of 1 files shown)


src/trace/trigger.ts
```diff
@@ -303,7 +303,7 @@ export function extractTriggerTags(event: any, context: Context) {
     try {
       eventSourceARN = parseEventSourceARN(eventSource, event, context);
     } catch (error) {
-      logError(`failed to extract ${eventSource} arn from the event`, { error });
+      logDebug(`failed to extract ${eventSource} arn from the event`, { error });
```
tianchu, line 306: BTW, is there any problem with the way we log `{ error }`? From both user reports, I didn't see any error being logged, which is probably the most valuable part of this log.

IvanTopolcic: Yeah this is actually super confusing because looking at the code all it seems like we're doing is returning a field in an object and somehow an error is being thrown. I've tried to replicate it [here](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgMJysg3gKGf5EAVwFsB1ACwggBsBnALkNICNoBuHAXxxxiJAIwwAPYhkAcwhhK1egAoEGJuigBKJsRJtMuAsijSiUcUqgA6LbNp1OPHAjF0whEWFXIAvNns4wUAE9sPHwpGSobeRA3VTVuZCUwBApkeQg1YP1HEDoRGghzGhEJeQAiAFEADyQAB2ExZDAKKBEAdxBSuKynPIKikvT4mFA4GhogvQJs3PzC4rKAMQBJADkAQQAZDYBNTu5eIA).

IvanTopolcic:
Change error log to debug

PR #300, opened 2022-08-22
aj/authorizer trace context

astuyve:
Fixes https://github.com/DataDog/serverless-plugin-datadog/issues/252 for NodeJS

### What does this PR do?
Encodes and decodes trace context from an authorizer function

### Motivation

Trace propagation and inferred span creation for token-based authorizers:
image

Trace propagation and inferred span creation for request-based authorizers:
image

### Testing Guidelines

Manual and automated testing

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


src/trace/context.ts
```diff
@@ -640,3 +546,22 @@ export function convertToAPMParentID(xrayParentID: string): string | undefined {
   }
   return hex.toString(10);
 }
+
+function exportTraceData(traceData: any): TraceContext | undefined {
```
DarcyRaynerDD, line 550: Nice refactor

src/trace/listener.ts
```diff
@@ -150,6 +157,17 @@ export class TraceListener {
     return false;
   }
 
+  injectAuthorizerSpan(result: any): any {
+    const finishTime = this.inferredSpan?.isAsync() ? this.wrappedCurrentSpan?.startTime() : Date.now();
```
DarcyRaynerDD, line 161: This logic is a bit hard to follow. Maybe explain what's going on in a comment

src/trace/span-inferrer.ts
```diff
@@ -52,6 +55,10 @@ export class SpanInferrer {
     return "sync";
   }
 
+  isTracedAuthorizerInvocation(event: any): boolean {
+    return event.requestContext?.authorizer?._datadog;
```
DarcyRaynerDD, line 59: Is there any danger from not explicitly coalescing this into a bool?

DarcyRaynerDD (approved):
Left some minor comments

src/trace/span-inferrer.ts
astuyve: It could evaluate true if the key `_datadog` is present but empty, or an empty string or something - but the next call is wrapped in a try block anyway (to handle the said case), so I think it's okay for the usage as is - but I'll add an explicit check for `!== undefined`, in case someone re-uses it in the future.

src/trace/span-inferrer.ts
```diff
@@ -97,9 +104,27 @@ export class SpanInferrer {
...
+        upstreamSpanOptions = {
+          startTime: parsedUpstreamContext[parentSpanFinishTimeHeader],
+          tags: { operation_name: "aws.apigateway.authorizer", ...options.tags },
+        };
+        upstreamSpanOptions.childOf = parentSpanContext;
+        upstreamAuthorizerSpan = new SpanWrapper(
+          this.traceWrapper.startSpan("aws.apigateway.authorizer", upstreamSpanOptions),
+          { isAsync: false },
+        );
+        upstreamAuthorizerSpan.finish(event.requestContext.timeEpoch);
```
joeyzhao2018, line 121: The `event.requestContext.timeEpoch` header exists for HTTP API (API Gateway V2).  
For REST API (API Gateway V1), the corresponding key is `event.requestContext.requestTimeEpoch`.
I also found out that `integrationLatency` is only available for REST API. 
Moreover, for HTTP API the key actually becomes  `event.requestContext.authorizer.lambda._datadog` [example](https://github.com/DataDog/datadog-lambda-python/blob/joey/authorizer_spans/tests/event_samples/authorizer-request-api-gateway-v2.json#L25-L26). So this current code here fits mostly with REST API. 

Also, in both cases, the epoch time here is the original user invocation timestamp. Therefore, I was using `the request time epoch + integrationlatency` to get an approximation in [the python version](https://github.com/DataDog/datadog-lambda-python/blob/joey/authorizer_spans/datadog_lambda/tracing.py#L712-L720). Please let me know what do you think.

In short, my suggestion for this line is
`upstreamAuthorizerSpan.finish(event.requestContext.requestTimeEpoch + event.requestContext.authorizer.integrationLatency);`

event_samples/api-gateway-traced-authorizer.json
```diff
@@ -0,0 +1,89 @@
...
+    "resourcePath": "/hello",
+    "httpMethod": "POST",
+    "extendedRequestId": "XIIseElXGjQFvXg=",
+    "requestTime": "19/Aug/2022:20:10:55 +0000",
+    "path": "/dev/hello",
+    "accountId": "601427279990",
+    "protocol": "HTTP/1.1",
+    "stage": "dev",
+    "domainPrefix": "3gsxz7lha4",
+    "requestTimeEpoch": 1660939855656,
```
joeyzhao2018, line 68: This is an example about the other comment

duncanista (commented):
Looks really great!

duncanista (approved):

PR #308, opened 2022-10-04
Add alternative method of handler redirection

DarcyRaynerDD:
### What does this PR do?

Adds a simpler handler redirection method. Now to apply the handler, the user just has to set this environment variable:
```
NODE_OPTIONS=-r datadog-lambda-js/init
```
This removes the need to set DD_LAMBDA_HANDLER, or override the handler on the function configuration.

### Motivation

Makes it easier to user the handler with other middle-ware doing handler indirection. 

### Testing Guidelines

I verified this manually on esm and nodejs.  This can be an experimental option until we are able to update our build tools.

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


scripts/update_dist_version.sh
```diff
@@ -17,4 +17,5 @@ echo "$TRACE_CONSTANTS" |
 
 echo "Copying handler js files"
 cp src/handler.* dist/
+cp src/init.js dist/init.js
```
astuyve, line 20: We also copy the handler(s) over in the Docker file, do we need to repeat that here? Or is the postbuild script enough?

astuyve (approved):

scripts/update_dist_version.sh
DarcyRaynerDD: The docker file has special handling for .cjs files. Since node 12 doesn't support .mjs, and doesn't recognize the .cjs extension, it has to be renamed. We can delete the docker lines, when we drop node12, and just have the .mjs entrypoint.

PR #331, opened 2023-01-04
Extract Lambda Trace Context from the Event's Step Functions Context

kimi-p:
### What does this PR do?
Set Lambda span's `trace_id` and `parent's span_id` when Step Functions Context is detected in the event. The corresponding Lambda subtask span is set in the backend https://github.com/DataDog/logs-backend/pull/43234.

The MD5 logic on the backend is here: https://github.com/DataDog/logs-backend/blob/329f0ce114138067af8cdaef82bdbe15a06ffc58/domains/serverless/apps/logs-to-traces-reducer/src/main/java/com/dd/logs/serverless/logs_to_traces/common/HashHelper.java#L12-L32

Node: md5 is used to generate a string to be converted to trace_id in a deterministic way, so md5 is not used for any encryption purposes.

### Motivation
To link Step Functions log generated span (parent) with Lambda span (child). 

### Testing Guidelines
Tested in staging with custom made dd-lambda-js layer. The trace is [here](https://dd.datad0g.com/apm/traces?query=%40_top_level%3A1%20env%3Astaging%20resource_name%3Alogs-to-traces-sequential%20OR%20resource_name%3ATestLambdaSpanLinking&cols=core_service%2Ccore_resource_name%2Clog_duration%2Clog_http.method%2Clog_http.status_code&graphType=flamegraph&historicalData=true&messageDisplay=inline&sort=desc&spanID=3797380583832534858&spanType=service-entry&spanViewType=metadata&timeHint=1673384121610&trace=AgAAAYWdd1UKP7jGWAAAAAAAAAAYAAAAAEFZV2RkMnZCQUFBdjQtUnpJbUttNXNoSQAAACQAAAAAMDE4NTlkN2ItNmY3Yy00ZDIxLTkzMjMtNjJiYWQ1ZTZmMzM2&traceID=5244573283293107403&start=1673370956921&end=1673385356921&paused=false).

image

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


src/trace/context.ts
```diff
@@ -108,6 +151,14 @@ export function extractTraceContext(
       }
     }
     return trace;
+  } else {
+    // do not send step functions context to xray
+    if (stepFuncContext) {
```
joeyzhao2018, line 156: nitpik: can we combine this with the block above in order to put the `stepFuncContext` related logic together. that is:
```javascript
  if (stepFuncContext) {
    try {
      addStepFunctionContextToXray(stepFuncContext);
    } catch (error) {
      if (error instanceof Error) {
        logError("couldn't add step function metadata to xray", error as Error);
      }
    }
    if (!trace) {
      trace = readTraceFromStepFunctionsContext(stepFuncContext);
      if (trace !== undefined) {
        return trace;
      }
    }
  }
```

joeyzhao2018 (approved):

src/trace/context.ts
kimi-p: Great suggestion. Updated and thanks for the review!

src/trace/context.ts
```diff
@@ -55,6 +56,48 @@ export interface StepFunctionContext {
   "step_function.state_retry_count": number;
 }
 
+export function readTraceFromStepFunctionsContext(stepFunctionContext: StepFunctionContext): TraceContext | undefined {
+  return {
+    traceID: deterministicMd5HashToBigIntString(stepFunctionContext["step_function.execution_id"]),
```
DarcyRaynerDD, line 61: So, this is an issue with the existing code, (which I wrote), but I don't like how are are indexing an object instead of using property names, I feel like it's bad javascript style. I think it was done that way originally so it would be easy to add as metadata to the APM span, (we have to use the "step_function." prefix so it's grouped in the UI) . Not a blocker for merging, but maybe an opportunity for refactoring.

src/trace/context.ts
```diff
@@ -55,6 +56,48 @@ export interface StepFunctionContext {
...
+  };
+}
+
+export function hexToBinary(hex: string) {
+  // convert hex to binary and padding with 0 in the front to fill 128 bits
+  return parseInt(hex, 16).toString(2).padStart(4, "0");
+}
+
+export function deterministicMd5HashInBinary(s: string): string {
+  // Md5 here is not used as an encryption method but to generate a deterministic hash as the backend does
```
DarcyRaynerDD, line 80: Pedantic edit, since hashing != encryption
```suggestion
  // Md5 here is used here because we don't need a cryptographically secure hashing method but to generate the same trace/span ids as the backend does
```

src/trace/context.spec.ts
```diff
@@ -1109,3 +1146,141 @@ describe("extractTraceContext", () => {
     `);
   });
 });
+
+describe("test_hexToBinary", () => {
```
DarcyRaynerDD, line 1150: nit: Might be worth using a table driven test here to cut down on boilerplate: 
https://jestjs.io/docs/api#describeeachtablename-fn-timeout

DarcyRaynerDD (approved):
Overall looks good. Nice work.

PR #372, opened 2023-03-28
feat: Modify end time of lambda load span to be last cold start node …

astuyve:
…end time instead of start time of function.

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/trace/cold-start-tracer.ts
+++ b/src/trace/cold-start-tracer.ts
@@ -6,6 +6,7 @@ export interface ColdStartTracerConfig {
   tracerWrapper: TracerWrapper;
   parentSpan?: SpanWrapper;
   lambdaFunctionName?: string;
+  currentSpanStartTime: number;
   minDuration: number;
   ignoreLibs: string;
 }
@@ -14,6 +15,7 @@ export class ColdStartTracer {
   private tracerWrapper: TracerWrapper;
   private parentSpan?: SpanWrapper;
   private lambdaFunctionName?: string;
+  private currentSpanStartTime: number;
   private minDuration: number;
   private ignoreLibs: string[];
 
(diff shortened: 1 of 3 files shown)


src/trace/cold-start-tracer.ts
```diff
@@ -15,28 +14,27 @@ export class ColdStartTracer {
...
     this.lambdaFunctionName = coldStartTracerConfig.lambdaFunctionName;
-    this.coldStartSpanFinishTime = coldStartTracerConfig.coldStartSpanFinishTime;
     this.minDuration = coldStartTracerConfig.minDuration;
     this.ignoreLibs = coldStartTracerConfig.ignoreLibs.split(",");
   }
 
   trace(rootNodes: RequireNode[]) {
     const coldStartSpanStartTime = rootNodes[0]?.startTime;
-    const coldStartSpan = this.createColdStartSpan(coldStartSpanStartTime, this.parentSpan);
+    const coldStartSpanEndTime = rootNodes[rootNodes.length - 1]?.endTime;
```
joeyzhao2018, line 30: If a require node were created during customer's code because they dynamically loaded the module,  `rootNodes[rootNodes.length - 1].endTime` can potentially end up outside the outer cold-trace span? (I haven't tried this in NodeJS but python example is like [this](https://ddserverless.datadoghq.com/apm/trace/15416331496885506621?colorBy=service&env=dev&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=14889763467639317383&spanViewType=metadata&timeHint=1680046997360))

astuyve: That's a good point. Maybe we take the oldest timestamp, either the _last_ root node end time or the parent span start time?

PR #387, opened 2023-05-03
feat: Support async trace context extraction

### LINKED ISSUE #385: Make `TraceExtractor` async
We would like to trace out serverless application, which include the following scenario

1. A lambda function performs a POST operation to API Gateway.
2. On receive, a lambda is triggered taking the payload and persisting it to an S3 bucket, saving the trace context as metadata.
3. On this S3 bucket, an `Object Created` event is delivered to EventBridge.
4. A lambda is triggered on this event.

Our intention is to have a full trace of this flow showing up in the UI. Unfortunately, there is no possibility to pass metadata from the S3 object to EventBridge. AWS simply does not support it. Out idea was to utilize DD_TRACE_EXTRACTOR to perform a headObject operation and extracting the trace context, however, neither the function `extractTraceContext` nor `TraceExtractor`
in this library is async and therefore we can't use promises inside the extractor.

Looking at the code shows that, e.g., `MetricsListener` i already implemented to use async functions, which is not the case for `TraceListener`. Is there any specific reason for this?

astuyve:
### What does this PR do?
This PR permits a custom trace context extractor to be asynchronous

### Motivation
Fixes #385

### Testing Guidelines
I added a new unit test to cover the async path, and repurposed an existing test to cover the synchronous path.

### Additional Notes
I tested this manually by requesting `datadoghq.com` in an async extractor
```js
module.exports.json = async (payload) => {
  const res = await axios.get('https://datadoghq.com');
  console.log('GOT RES IN ASYNC HANDLER:', res);
  return;
};
```

We can see that the call was awaited and the result returned, even though the extractor returned null, the logged result from [this line](https://github.com/DataDog/datadog-lambda-js/blob/main/src/trace/context.ts#L116) is shown after the request finished:
image

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -138,7 +138,7 @@ export function datadog<TEvent, TResult>(
     currentTraceListener = traceListener;
 
     try {
-      traceListener.onStartInvocation(event, context);
+      await traceListener.onStartInvocation(event, context);
       await metricsListener.onStartInvocation(event);
       if (finalConfig.enhancedMetrics) {
         incrementInvocationsMetric(metricsListener, context);
--- a/src/trace/context.spec.ts
+++ b/src/trace/context.spec.ts
@@ -817,10 +817,10 @@ describe("extractTraceContext", () => {
     process.env["_X_AMZN_TRACE_ID"] = undefined;
     process.env[awsXrayDaemonAddressEnvVar] = undefined;
   });
-  it("returns trace read from header as highest priority with no extractor", () => {
+  it("returns trace read from header as highest priority with no extractor", async () => {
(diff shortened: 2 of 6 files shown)


sfirrin (approved):

midu-git:
@astuyve Many thanks for the very fast resolution resulting in this PR. FYI: I've checked out your branch and build the layer, and I've successfully tested it within my environment for the case described in the initial [issue](https://github.com/DataDog/datadog-lambda-js/issues/385).

astuyve:
Thanks @midu-git! We should be able to release this shortly

PR #390, opened 2023-05-09
Pass tracer explicitly to datadog wrapper to avoid side effects when

midu-git:
### What does this PR do?

The tracer is passed explicitly to `datadog` wrapper. This reduces the dynamic resolution of `dd-trace` to a single place, and resolves the following issue:

Assume that you use the libraries `dd-trace` and `datadog-lambda-js` as npm modules without the layer. Moreover, assume you use a bundler like `esbuild`. In this situation, when initializing the trace as follows

```
import { tracer } from "dd-trace"
tracer.init(...)

export handler = datadog(handler_to_be_wrapped, ...)
```

and bundling this code with `esbuild` say

```
esbuild.build({
    entryPoints: [above_file],
    bundle: true,
    treeShaking: true,
    minify: true,
    sourcemap: true,
    entryNames: "[name]/index",
    outdir: "dist",
    platform: "node",
    target: ["node18"],
    logLevel: "info",
    plugins: [require("dd-trace/esbuild")],
})
```

the trace initialized will not be the same as the one this library loads, resulting in using a noop tracer for calls within this library. Up to my knowledge, there is no possibility to resolve this issue otherwise.

For the layer case, this change will not have an impact. However, the `datadog` wrapper has now a mandatory additional attribute. Thus, this change is potentially a breaking change.

To minimize the impact, loading and initializing the tracer has been split into two different functions, the latter one being exported to the user. Thus the user is able to initialize the tracer exactly the same way this library would do it.

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [x] Breaking change

### Check all that apply

### DIFF
--- a/src/handler.cjs
+++ b/src/handler.cjs
@@ -7,15 +7,17 @@ const {
 } = require("./index.js");
 const { logDebug, logError } = require("./utils/index.js");
 const { loadSync } = require("./runtime/index.js");
-const { initTracer } = require("./runtime/module_importer");
+const { loadTracer } = require("./runtime/module_importer.js");
+const { initTracer } = require("./trace/index.js");
 
 if (process.env.DD_TRACE_DISABLED_PLUGINS === undefined) {
   process.env.DD_TRACE_DISABLED_PLUGINS = "fs";
   logDebug("disabled the dd-trace plugin 'fs'");
 }
 
+const tracer = loadTracer();
 if (getEnvValue("DD_TRACE_ENABLED", "true").toLowerCase() === "true") {
-  initTracer();
+  initTracer(tracer);
 }
 
 const taskRootEnv = getEnvValue(lambdaTaskRootEnvVar, "");
@@ -32,4 +34,4 @@ if (extractorEnv) {
   }
 }
 
-exports.handler = datadog(loadSync(taskRootEnv, handlerEnv), { traceExtractor });
+exports.handler = datadog(loadSync(taskRootEnv, handlerEnv), tracer, { traceExtractor });
--- a/src/handler.mjs
+++ b/src/handler.mjs
@@ -1,15 +1,17 @@
 import { datadog, datadogHandlerEnvVar, lambdaTaskRootEnvVar, traceExtractorEnvVar, getEnvValue } from "./index.js";
 import { logDebug, logError } from "./utils/index.js";
 import { load } from "./runtime/index.js";
-import { initTracer } from "./runtime/module_importer.js";
+import { loadTracer } from "./runtime/module_importer.js";
+import { initTracer } from "./trace/index.js";
 
 if (process.env.DD_TRACE_DISABLED_PLUGINS === undefined) {
   process.env.DD_TRACE_DISABLED_PLUGINS = "fs";
   logDebug("disabled the dd-trace plugin 'fs'");
 }
 
+const tracer = loadTracer();
 if (getEnvValue("DD_TRACE_ENABLED", "true").toLowerCase() === "true") {
-  initTracer();
+  initTracer(tracer);
 }
 
 const taskRootEnv = getEnvValue(lambdaTaskRootEnvVar, "");
@@ -26,4 +28,4 @@ if (extractorEnv) {
   }
 }
 
-export const handler = datadog(await load(taskRootEnv, handlerEnv), { traceExtractor });
+export const handler = datadog(await load(taskRootEnv, handlerEnv), tracer, { traceExtractor });
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -13,6 +13,7 @@ import { incrementErrorsMetric, incrementInvocationsMetric } from "./metrics/enh
 import { LogLevel, setLogLevel } from "./utils";
 import { HANDLER_STREAMING, STREAM_RESPONSE } from "./constants";
 import { PassThrough } from "stream";
+import { loadTracer } from "./runtime/module_importer.js";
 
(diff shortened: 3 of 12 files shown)


midu-git:
Hi guys. Is there anything I need to do in order for the workflows to be started?

astuyve:
Hi @midu-git - thanks for reaching out with this PR. I'm a bit confused on your example.

By chance are you also redirecting your handler to datadog _and_ wrapping it with the `datadog` method?

If you are following [these steps](https://docs.datadoghq.com/serverless/guide/handler_wrapper/), you do not need to also redirect your lambda function handler. I think that may be the confusion here. 

This code works in my test example:
```js
'use strict';
const { datadog } = require('datadog-lambda-js');
const tracer = require("dd-trace").init({});

module.exports.hello = datadog( async (event) => {
  return {
    statusCode: 200,
    body: JSON.stringify(
      {
        message: 'Go Serverless v1.0! Your function executed successfully!',
        input: event,
      },
      null,
      2
    ),
  };

  // Use this code if you don't use the http event with the LAMBDA-PROXY integration
  // return { message: 'Go Serverless v1.0! Your function executed successfully!', event };
});
```
as demonstrated by this trace:
image

If possible, could you provide an example which reproduces this issue so we can understand this case?

Thanks!

midu-git:
Hi @astuyve. As mentioned, my setup involves a bundler. In particular esbuild. In this case it seems like two code paths are being used. I'll prepare a setup similar to yours, including bundling.

Many thanks so far.

midu-git:
Hi @astuyve.
I've prepared a small setup including
* lambda as prepared by you
* esbuild for bundling
* pulumi for deploying the lambda function

You can find the code [here](https://github.com/midu-git/dd-lambda-bundler). Here is the debug output of a run [log-events-viewer-result.csv](https://github.com/DataDog/datadog-lambda-js/files/11495898/log-events-viewer-result.csv).
Clearly, the agent starts, but the tracer used by datadog-lambda-js seems to be not the one imported and initialized in the handler function, and thus being a noop tracer.
Of course one can argue that this might be a bug within esbuild. However, my approach will reduce the dynamic loading to a single place, and thus provides a less implicit behaviour.

Hopefully this example clarifies the issue I'm facing.

BR

astuyve:
Hi @midu-git - thanks again for your help with this reproduction case. I'll ask an engineer to look into it, but I suspect it will take some time as I'm not sure anyone is familiar with Pulumi on our end and we'll be off-site next week as a team.

I believe what is occurring is that `ddplugin` is also loading and initializing [ddtrace](https://github.com/DataDog/dd-trace-js/blob/master/esbuild.js#L3).

If you remove the [DD plugin](https://github.com/midu-git/dd-lambda-bundler/blob/main/scripts/build.js#L3), does that resolve this? I feel like you could remove either the initialization in your code or ddplugin and test.

Thanks!

midu-git:
Hi @astuyve.

1. Removing the plugin does not change the bundled code, in this case at least. Moreover, removing the plugin might not be an option, as in that case some interception mechanism might be missing.
2. Pulumi is just used for deployment. I could have used serverless or terraform or any other mechanism. I have used it since we use it in our project and I didn't want to set up a new deployment procedure.

However, what is wrong with the basic idea provided in this PR?

BR

astuyve:
Okay thanks for that context. I'm asking a lot of questions because many of our users (myself included) also use esbuild and as far as I know this specifically has not been an issue, which is why it feels like a configuration issue.

Additionally this is compounded by the fact that this PR is a breaking change, which is not something we could release for some time.

I'll be out of the office next week on vacation, but will assign someone to look and follow up.

Thanks!

duncanista:
Hey – just checking in that I am investigating this issue. I will come up with more updates during the week.

duncanista:
Hey, I was able to reproduce and narrow down the problem.

First of all, I decided not to use Pulumi. If I could get it working, that would be my number one culprit. Therefore, using any other framework is perfectly alright for this situation. I used the serverless framework. How I deployed doesn't matter.

I've also decided to remove initializing the tracer (again) in the file. Just to understand if initializing twice would cause an error. This was not an issue at all. 

At the end, I found out that disabling `dd-trace/esbuild` made everything work as expected, instead of using the plugin, I just specified the packages to externalize, same as if I was using `esbuild-node-externals`.

I have yet to find the direct culprit inside `dd-trace-js`, but I hope this information is useful in the meantime. Will keep researching and I'll keep you posted.

– on the PR side, I don't think I'm optimistic in merging this change. Since it would be a breaking change and it's not the design we are looking for. Once I find the issue with the tracer, I'll contact the team responsible for it and hopefully we will work on something to fix this up.

I'd recommend following this approach until we can fix it:
https://docs.datadoghq.com/serverless/guide/serverless_tracing_and_bundlers/

duncanista:
Hey @midu-git,

Could you add the following to your `esbuild.config.js`?

```js
esbuild.build({
   entryPoints: [sourceFile],
   external: ["aws-sdk", "datadog-lambda-js", "dd-trace"], // <- this actually doesn't remove them from your build, it just externalizes them and keeps them inside of `node_modules`
   bundle: true,
   plugins: [ddPlugin],
})
```

And let me know if that worked for you? It seems to have fixed it in my proof of concept.
We normally tell the users to exclude them since they are provided in our Lambda Layers and they are not bundled.

I'd appreciate if you could fill a support ticket so we can prioritize making `datadog-lambda-js` an ESModule and make it work out of the box without having to externalize it.

duncanista:
Closing due to inactivity – feel free to reopen the issue if needed.

PR #392, opened 2023-05-11
DD_SERVICE_MAPPING for inferred spans

zARODz11z:
### What does this PR do?
allows users to remap the service name on inferred spans created in dd lambda js. Pairs with https://github.com/DataDog/documentation/pull/18144

### Motivation
SLS-3501

### Testing Guidelines
unit

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/trace/span-inferrer.spec.ts
+++ b/src/trace/span-inferrer.spec.ts
@@ -57,12 +57,17 @@ describe("SpanInferrer", () => {
     });
   });
 
-  it("remaps SNS inferred span service name based on DD_SERVICE_MAPPING", () => {
-    process.env.DD_SERVICE_MAPPING = "sns:new-name";
+  it("remaps all SNS inferred span service name based on DD_SERVICE_MAPPING", () => {
+    // Set the environment variable that depicts test outcome, i.e. the result of service tag
+    process.env.DD_SERVICE_MAPPING = "sns|new-name";
     const inferrer = new SpanInferrer(mockWrapper as unknown as TracerWrapper);
     inferrer.createInferredSpan(snsEvent, {} as any, {} as SpanContext);
 
-    expect(mockWrapper.startSpan).toBeCalledWith("aws.sns", {
+    let modifiedSnsEvent = JSON.parse(JSON.stringify(snsEvent));
+    modifiedSnsEvent.Records[0].EventSubscriptionArn = "arn:aws:sns:us-east-1:123456789012:DifferentTopic";
+    inferrer.createInferredSpan(modifiedSnsEvent, {} as any, {} as SpanContext);
+
+    expect(mockWrapper.startSpan).toHaveBeenNthCalledWith(1,"aws.sns", {
       childOf: {},
       startTime: 1643039127968,
       tags: {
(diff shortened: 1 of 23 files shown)


src/trace/span-inferrer.ts
```diff
@@ -333,14 +356,16 @@ export class SpanInferrer {
...
     const resourceName = queueName;
+    const serviceName = this.serviceMapping.sqs || "sqs";
+
     options.tags = {
       operation_name: "aws.sqs",
       resource_names: resourceName,
       request_id: context?.awsRequestId,
       "span.type": "web",
       "resource.name": resourceName,
       "service.name": resourceName,
```
zARODz11z, line 367: @astuyve Whats service.name used for? As opposed to service?

src/trace/span-inferrer.ts
```diff
@@ -62,6 +64,20 @@ export class SpanInferrer {
     return "sync";
   }
 
+  getServiceMapping(): Record<string, string> {
```
astuyve, line 67: could be static, no?

src/trace/span-inferrer.ts
```diff
@@ -80,6 +96,8 @@ export class SpanInferrer {
       method = event.requestContext.http.method;
     }
     const resourceName = [method || domain, resourcePath].join(" ");
+    const serviceName = this.serviceMapping.api_gateway || domain;
```
astuyve, line 99: Your implementation here is correct, but I'm curious about the overlap with dd-trace-js using the same ENV var for service name remapping.

If I have a function triggered by API Gateway, but my lambda function also makes a request to any API Gateway endpoint, seems (or sqs, sns, etc), we need to make sure we don't accidentally overwrite the downstream spans.

Potentially we can get around this with a prefix like `lambda.apiGateway`

astuyve (changes requested):
Let's chat with @rochdev and make sure we're in harmony 🎶 with dd-trace before merging, but the implementation here looks really good. Nicely done Andrew!

src/trace/span-inferrer.ts
zARODz11z: We remap based on plugin names, meaning
` DD_SERVICE_MAPPING: "aws-sdk:my-new-service-name"`

on a lambda such as:

```
'use strict';
const AWS = require('aws-sdk');
const sqs = new AWS.SQS({region: 'sa-east-1'}); // Instantiate the SQS service client

module.exports.handler = async (event, context) => {
  try {
    // Send a message to an SQS queue
    const sqsParams = {
      QueueUrl: 'foobar', // Replace with your Queue URL
      MessageBody: 'Information about the current operation'
    };
    const sqsResponse = await sqs.sendMessage(sqsParams).promise();

    return {
      statusCode: 200,
      body: JSON.stringify({
        apiGatewayResponse: apiGatewayResponse.items,
        sqsResponse
      }),
    };
  } catch (err) {
    return {
      statusCode: 500,
      body: JSON.stringify(err),
    };
  }
};

```

has this effect

Screen Shot 2023-05-11 at 5 12 51 PM

So the prefix shouldn't be necessary

zARODz11z: Meaning our implementation of renaming inferred spans service names should not rename any service names downstream.

src/trace/span-inferrer.spec.ts
```diff
@@ -28,6 +28,10 @@ describe("SpanInferrer", () => {
     mockWrapper.startSpan.mockClear();
```
zARODz11z, line 28: Added documentation https://github.com/DataDog/documentation/pull/18144

src/trace/span-inferrer.spec.ts
```diff
@@ -53,6 +57,107 @@ describe("SpanInferrer", () => {
...
   });
 
+  it("remaps all SNS inferred span service name based on DD_SERVICE_MAPPING", () => {
+    // Set the environment variable that depicts test outcome, i.e. the result of service tag
+    process.env.DD_SERVICE_MAPPING = "lambda_sns:new-name";
+    const inferrer = new SpanInferrer(mockWrapper as unknown as TracerWrapper);
+    inferrer.createInferredSpan(snsEvent, {} as any, {} as SpanContext);
+
+    let modifiedSnsEvent = JSON.parse(JSON.stringify(snsEvent));
+    modifiedSnsEvent.Records[0].EventSubscriptionArn = "arn:aws:sns:us-east-1:123456789012:DifferentTopic";
```
zARODz11z, line 67: instead of creating a whole new event we can just edit some characteristics to simulate a different even

src/trace/span-inferrer.spec.ts
```diff
@@ -53,6 +57,107 @@ describe("SpanInferrer", () => {
...
+
+  it("remaps specific SNS inferred span service name based on DD_SERVICE_MAPPING topicname", () => {
+    // Set the environment variable that depicts test outcome, i.e. the result of service tag
+    process.env.DD_SERVICE_MAPPING = "DifferentTopic:new-name";
+    console.log(process.env.DD_SERVICE_MAPPING);
+    const inferrer = new SpanInferrer(mockWrapper as unknown as TracerWrapper);
+    inferrer.createInferredSpan(snsEvent, {} as any, {} as SpanContext);
+
+    let modifiedSnsEvent = JSON.parse(JSON.stringify(snsEvent));
+    modifiedSnsEvent.Records[0].Sns.TopicArn = "arn:aws:sns:us-east-1:123456789012:DifferentTopic";
```
zARODz11z, line 118: Causes a different topicname to be extracted here https://github.com/DataDog/datadog-lambda-js/pull/392/files#diff-c90b7ab89897f78a5515f4fbe787c8b74912175b75a69a58459a1bec936f2498R299

thus we can assert below only the service associated with the topic name `DifferentTopic` is remapped to new-name

src/trace/span-inferrer.spec.ts
```diff
@@ -53,6 +57,107 @@ describe("SpanInferrer", () => {
...
+      startTime: 1643039127968,
+      tags: {
+        _inferred_span: { synchronicity: "async", tag_source: "self" },
+        event_subscription_arn: "arn:aws:sns:us-east-1:123456789012:ExampleTopic",
+        message_id: "95df01b4-ee98-5cb9-9903-4c221d41eb5e",
+        operation_name: "aws.sns",
+        request_id: undefined,
+        "resource.name": "DifferentTopic",
+        resource_names: "DifferentTopic",
+        service: "new-name",
```
zARODz11z, line 151: ^ new-name caused by 
process.env.DD_SERVICE_MAPPING = "DifferentTopic:new-name"

astuyve (approved):
Great job on this Andrew, way to push it through!

src/trace/span-inferrer.ts
```diff
@@ -15,11 +15,29 @@ import { parentSpanFinishTimeHeader } from "./constants";
...
+  private static initServiceMapping() {
+    const serviceMappingStr = process.env.DD_SERVICE_MAPPING || "";
+
+    serviceMappingStr.split(",").forEach((entry) => {
+      const [key, value] = entry.split(":");
+      if (key && value) {
+        this.serviceMapping[key.trim()] = value.trim();
+      }
+    });
+  }
```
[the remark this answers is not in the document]
zARODz11z: yeah, I added a few more test cases

duncanista (approved):
LGTM!

PR #402, opened 2023-07-07
feat: Support kms encrypted API keys for creating distribution metric…

### LINKED ISSUE #395: Full aws-sdk v3 support for nodejs18.x - remove v2 leftovers
## Expected Behavior

No aws-sdk v2 dependencies

## Actual Behavior

Still hidden (as in: not obvious from package.json dependencies) aws-sdk v2 dependencies - e.g. [kms-service](https://github.com/DataDog/datadog-lambda-js/blob/main/src/metrics/kms-service.ts) depends on `aws-sdk/clients/kms`. The source code comment and rationale
```
// in the Lambda environment anyway), we use require to import the SDK.
```
is since nodejs18.x not correct anymore. So basically, if I want to use this library in a lambda with nodejs18.x runtime, I need to bundle in the whole aws-sdk-v2 code into the lambda code. 

## Specifications

  - Datadog Lambda Layer version: 7.9.x
  - Node version: 18

## Stacktrace

  ```
 ../app/node_modules/datadog-lambda-js/dist/metrics/kms-service.js:54:38:
      54 │                         kms = require("aws-sdk/clients/kms");
         ╵                                       ~~~~~~~~~~~~~~~~~~~~~

  ```

astuyve:
…s directly using aws sdk v3 in node18

### What does this PR do?

Fixes #395 

### Motivation
Users directly sending custom metrics from Lambda using node 18 previously had to supply their own version of the kms client from the aws sdk v2.

This code honors both v2 sdk or v3 sdk. We can see in node 18, the v3 kms client is loaded:
image

and I confirmed metrics are sent properly:
image

### Testing Guidelines

Manually as well as unit tests.

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/metrics/kms-service.ts
+++ b/src/metrics/kms-service.ts
@@ -1,14 +1,19 @@
 // In order to avoid the layer adding the 40mb aws-sdk to a deployment, (which is always available
 // in the Lambda environment anyway), we use require to import the SDK.
 
-import { logError } from "../utils";
-
 export class KMSService {
+  private encryptionContext
+
+  constructor () {
+    this.encryptionContext = { LambdaFunctionName: process.env.AWS_LAMBDA_FUNCTION_NAME ?? "" };
+  }
+
   public async decrypt(ciphertext: string): Promise<string> {
+    const buffer = Buffer.from(ciphertext, "base64");
+
     try {
       const kms = require("aws-sdk/clients/kms");
       const kmsClient = new kms();
-      const buffer = Buffer.from(ciphertext, "base64");
 
       // When the API key is encrypted using the AWS console, the function name is added as an encryption context.
       // When the API key is encrypted using the AWS CLI, no encryption context is added.
@@ -19,8 +24,7 @@ export class KMSService {
         result = await kmsClient.decrypt({ CiphertextBlob: buffer }).promise();
       } catch {
         // Then try with encryption context, in case API key was encrypted using the AWS Console
-        const encryptionContext = { LambdaFunctionName: process.env.AWS_LAMBDA_FUNCTION_NAME ?? "" };
-        result = await kmsClient.decrypt({ CiphertextBlob: buffer, EncryptionContext: encryptionContext }).promise();
+        result = await kmsClient.decrypt({ CiphertextBlob: buffer, EncryptionContext: this.encryptionContext }).promise();
       }
 
       if (result.Plaintext === undefined) {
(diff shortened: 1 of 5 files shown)


src/metrics/kms-service.ts
```diff
@@ -29,11 +35,26 @@ export class KMSService {
...
+  public async decryptV3(buffer: Buffer): Promise<string> {
+    const { KMSClient, DecryptCommand } = require("@aws-sdk/client-kms");
+    const kmsClient = new KMSClient();
+    let result;
+    try {
+      const decryptCommand = new DecryptCommand({ CiphertextBlob: buffer });
+      result = await kmsClient.send(decryptCommand);
+    } catch {
+      const decryptCommand = new DecryptCommand({ CiphertextBlob: buffer, EncryptionContext: this.encryptionContext });
+      result = await kmsClient.send(decryptCommand);
```
duncanista, line 55: Is it guaranteed that this will never fail?

src/metrics/kms-service.ts
```diff
@@ -19,8 +24,9 @@ export class KMSService {
         result = await kmsClient.decrypt({ CiphertextBlob: buffer }).promise();
       } catch {
         // Then try with encryption context, in case API key was encrypted using the AWS Console
-        const encryptionContext = { LambdaFunctionName: process.env.AWS_LAMBDA_FUNCTION_NAME ?? "" };
-        result = await kmsClient.decrypt({ CiphertextBlob: buffer, EncryptionContext: encryptionContext }).promise();
+        result = await kmsClient
+          .decrypt({ CiphertextBlob: buffer, EncryptionContext: this.encryptionContext })
+          .promise();
```
duncanista, line 29: Same for this? I wonder if this code somehow would fail and throw something from our side?

duncanista (approved):
LGTM, left a comment!

src/metrics/kms-service.ts
astuyve: It always could throw, we catch and log here, where we don't re-raise https://github.com/DataDog/datadog-lambda-js/blob/main/src/metrics/listener.ts#L185

duncanista: I missed that, thank you!

src/metrics/kms-service.ts
astuyve: no, but my above comment applies - this logic would raise through `decrypt` which would then continue.

I think I will modify the logic to always log, regardless of the error class though.

PR #412, opened 2023-08-04
feat: eventbridge to sqs support

astuyve:
### What does this PR do?

Adds support for tracing event bridge to sqs payloads

### Motivation

image

### Testing Guidelines

Units should be okay, I tested manually and shared a screenshot above.

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


src/trace/span-inferrer.ts
```diff
@@ -355,6 +355,37 @@ export class SpanInferrer {
     return new SpanWrapper(this.traceWrapper.startSpan("aws.sns", options), spanWrapperOptions);
   }
 
+  createInferredSpanForEBSQS(
```
duncanista, line 358: I think this method is not needed, since `createInferredSpanForEventBridge` diff is only in the signature type for `event`, and of course the name.

event_samples/eventbridge-sqs.json
```diff
@@ -0,0 +1 @@
...roPEJ5R5nwhLOEbeN5HROlZRXeaAwZOIN8BjqdeooYTIOrtvMEVb7a6OPLMdH1XB+ddevtKAH8K9Tm2ZjpaA7dtBGh1zFVHzBk=","body":"{\\"version\\":\\"0\\",\\"id\\":\\"af718b2a-b987-e8c0-7a2b-a188fad2661a\\",\\"detail-type\\":\\"my.Detail\\",\\"source\\":\\"my.Source\\",\\"account\\":\\"425362996713\\",\\"time\\":\\"2023-08-03T22:49:03Z\\",\\"region\\":\\"us-east-1\\",\\"resources\\":[],\\"detail\\":{\\"text\\":\\"Hello, world!\\",\\"_datadog\\":{\\"x-datadog-trace-id\\":\\"7379586022458917877\\",\\"x-datadog-parent-id\\":\\"2644033662113726488\\",\\"x-datadog-sampling-priority\\":\\"1\\",\\"x-datadog-tags\\":\\"_dd.p.dm=-0\\",\\"traceparent\\":\\"00-000000000000000066698e63821a03f5-24b17e9b6476c018-01\\",\\"tracestate\\":\\"dd=t.dm:-0;s:1\\"}}}","attributes":{"ApproximateReceiveCount":"1","AWSTraceHeader":"Root=1-64cc2edd-112fbf1701d1355973a11d57;Parent=7d5a9776024b2d42;Sampled=0","SentTimestamp":"1691102943638","SenderId":"AIDAJXNJGGKNS7OSV23OI","ApproximateFirstReceiveTimestamp":"1691102943647"},"messageAttributes":{},"md5OfBody":"93d9f0cd8886d1e000a1a0b7007bffc4","eventSource":"aws:sqs","eventSourceARN":"arn:aws:sqs:us-east-1:425362996713:lambda-eb-sqs-lambda-dev-demo-queue","awsRegion":"us-east-1"}]}
```
astuyve, line 1: TODO - format

src/trace/span-inferrer.ts
astuyve: Yeah I think you're right, same with the SNSSQS method.

event_samples/eventbridge-sqs.json
duncanista: This Payload doesn't seem to be EventBridge-SQS, unless I'm missing something, but I don't see any `Records` on your `body`.

```json
"body": {
    "version": "0",
    "id": "af718b2a-b987-e8c0-7a2b-a188fad2661a",
    "detail-type": "my.Detail",
    "source": "my.Source",
    "account": "425362996713",
    "time": "2023-08-03T22: 49: 03Z",
    "region": "us-east-1",
    "resources": [],
    "detail": {
        "text": "Hello, world!",
        "_datadog": {
        "x-datadog-trace-id": "7379586022458917877",
        "x-datadog-parent-id": "2644033662113726488",
        "x-datadog-sampling-priority": "1",
        "x-datadog-tags": "_dd.p.dm=-0",
        "traceparent": "00-000000000000000066698e63821a03f5-24b17e9b6476c018-01",
        "tracestate": "dd=t.dm: -0;s: 1"
        }
    }
},
```

astuyve: One must reverse their brain :D

This is a message received by a Lambda function, which was from an SQS queue, and before that, an Event Bus.

So `Records`, which is the first key of this JSON object, represents all the records of the SQS payload. The first item is a message, and that first message has a body, which is the EventBridge Event.

duncanista: Oh lol, I totally missed that, thank you.

astuyve: No worries, I have to cross my eyes to see it!

src/utils/event-type-guards.ts
```diff
@@ -83,6 +83,18 @@ export function isSNSSQSEvent(event: any): event is SQSEvent {
   return false;
 }
 
+export function isEBSQSEvent(event: any): event is SQSEvent {
+  if (Array.isArray(event.Records) && event.Records.length > 0 && event.Records[0].eventSource === "aws:sqs") {
+    try {
+      const body = JSON.parse(event.Records[0].body) as EventBridgeEvent<any, any>;
+      return body["detail-type"] !== undefined;
```
duncanista, line 90: I wonder if we should be checking `detail` instead of `detail-type`, since we are getting the data from there.

I point this out since this can be an optional property.

astuyve: totally, I need to fix that in the other method as well - great point.

astuyve: Actually I think it's required, right? See `Minimum information needed for a valid custom event`:
https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-events-structure.html

src/trace/span-inferrer.spec.ts
```diff
@@ -568,6 +569,55 @@ describe("SpanInferrer", () => {
...
+    expect(mockWrapper.startSpan.mock.calls).toEqual([
+      [
+        "aws.eventbridge",
+        {
+          childOf: {},
+          startTime: 1691102943000,
+          tags: {
+            _inferred_span: { synchronicity: "async", tag_source: "self" },
+            operation_name: "aws.eventbridge",
+            "peer.service": "mock-lambda-service",
```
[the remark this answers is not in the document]
astuyve: Nah, `peer.service` needs to be a real service (not inferred) - you can see that in the sns -> sqs tests which are in this project

src/utils/event-type-guards.ts
duncanista: I was guiding myself from https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-eventbridge/interfaces/puteventsrequestentry.html
Screenshot 2023-08-04 at 3 36 55 PM

But if we are really strict about it, `Detail` is also optional 😕

src/trace/span-inferrer.spec.ts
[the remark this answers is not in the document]
duncanista: Sounds good then!

duncanista (approved):

src/utils/event-type-guards.ts
astuyve: I don't think those docs are correct, I can try it in the console and verify but I'm pretty sure you need both

PR #429, opened 2023-11-07
[w3c] Support W3C trace context propagation

duncanista:
### What does this PR do?

TLDR; Refactor code and delegate trace extraction to the tracer.

This PR does a couple things:
1. Refactoring functions into classes to make trace context extraction simpler.
2. Delegate the trace context extraction to the tracer, instead of doing it manually only for Datadog headers.
3. Support W3C and Datadog headers extraction – actually, any supported by the current tracer.
4. Add multiple new unit tests for every component.
5. Separates Xray and Step Functions into services.

### Motivation

Our library was extracting tracing context only for Datadog headers, the need to extract from W3C trace context is expected and already being done by the `dd-trace` package. Therefore, we should delegate our previous manual extraction to the already predefined functions.

### Testing Guidelines

- [x] Unit tests
- [ ] Integration tests
- [x] Manual tests

#### Services using W3C context in headers in AWS Lambda.
Screenshot 2023-11-07 at 8 26 07 PM

Screenshot 2023-11-07 at 8 36 32 PM

### Additional Notes

- Patching for HTTP and `console` when no tracer is available will fallback to always propagate Datadog Headers only.
- Soft deprecation of `traceID` and `parentID` in the custom extractor expected value is in place, in the next breaking change, this will be removed in favor of `traceId` and `parentId`.

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


duncanista:
It's normal for integration tests to not pass, since now, console patching only happens when `dd-trace` is available, which is a contradiction – and a breaking change. Will need to fix it either by doing a try-catch, but this might require to handle scenarios where we can get trace-context when there's no tracer: Xray, StepFunctionContext, and Custom extractor.

src/trace/context/extractor.spec.ts
purple4reina, line 1: Big PR so I could be missing it, but is there a place where we test all these methods with `traceparent` and `tracestate` headers? I'm only seeing tests that mention `x-datadog-*` headers.

src/trace/context/extractors/http.spec.ts
```diff
@@ -0,0 +1,294 @@
...
+          priority: "2",
+        },
+      };
+      const tracerWrapper = new TracerWrapper();
+
+      const payload = {
+        headers: {
+          "X-Datadog-Parent-Id": "4726693487091824375",
+          "X-Datadog-Sampling-Priority": "2",
+          "X-Datadog-Trace-Id": "797643193680388254",
```
purple4reina, line 79: Yes, it seems to me that the extractor can properly handle header with different cases.

src/trace/xray-service.ts
```diff
@@ -0,0 +1,204 @@
...
+
+    return {
+      traceId,
+      parentId,
+      sampled,
+    };
+  }
+
+  private convertToSampleMode(xraySampled: number): SampleMode {
+    return xraySampled === 1 ? SampleMode.USER_KEEP : SampleMode.USER_REJECT;
```
purple4reina, line 107: We should check to see if this is the correct mapping. I set the mode to `AUTO_KEEP` when the sampled value is `1` in the extension. I believe I pulled that logic from the java tracer.

src/trace/xray-service.ts
```diff
@@ -0,0 +1,204 @@
...
+      },
+    });
+  }
+
+  private parseTraceContextHeader(): XrayTraceHeader | undefined {
+    if (this.header === undefined) return;
+
+    // Example: Root=1-5e272390-8c398be037738dc042009320;Parent=94ae789b969f1cc5;Sampled=1
+    logDebug(`Reading Xray trace context from env var ${this.header}`);
+    const [root, parent, _sampled] = this.header.split(";");
```
purple4reina, line 91: Careful! The xray trace header can have more than three keys in it!

purple4reina (changes requested):
One request, next time it will make reviewing a lot easier if this is split into multiple PRs. The first one can be just moving the existing code around to new files, the second one can then edit those files. As this is now, I have no way to know if this is new code that I'm reviewing or copy/pasted code.

src/trace/context/extractors/app-sync.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export class AppSyncEventTraceExtractor implements EventTraceExtractor {
+  private httpEventExtractor: HTTPEventTraceExtractor;
+
+  constructor(private tracerWrapper: TracerWrapper) {
+    this.httpEventExtractor = new HTTPEventTraceExtractor(this.tracerWrapper, false);
+  }
+
+  extract(event: any): SpanContextWrapper | null {
+    event.headers = event.request.headers;
```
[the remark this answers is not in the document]
duncanista: Yes, that is correct. It mutates the object – as far as I could tell from unit testing.

I think this is not ideal, but this is what the previous code, before refactoring, did. I imagine we should update this in the future. Either way, a more thorough test should be done here!

src/trace/context/extractors/http.ts
```diff
@@ -0,0 +1,95 @@
...
+      // need to set the trace context if using authorizer lambda in authorizing (non-cached) cases
+      try {
+        const eventSourceSubType: HTTPEventSubType = HTTPEventTraceExtractor.getEventSubType(event);
+        const injectedAuthorizerHeaders = HTTPEventTraceExtractor.getInjectedAuthorizerHeaders(
+          event,
+          eventSourceSubType,
+        );
+        if (injectedAuthorizerHeaders !== null) {
+          const _traceContext = this.tracerWrapper.extract(injectedAuthorizerHeaders);
+          if (_traceContext === null) return null;
```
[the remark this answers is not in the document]
duncanista: Maybe? I'm honestly just refactoring here with what we were previously doing, same as in your comment above for `AppSync`, maybe we can add a note and discuss this in the future.

src/trace/context/extractors/http.ts
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    const headers = event.headers;
+    const lowerCaseHeaders: { [key: string]: string } = {};
+
+    for (const key of Object.keys(headers)) {
+      lowerCaseHeaders[key.toLowerCase()] = headers[key];
+    }
+
+    const traceContext = this.tracerWrapper.extract(lowerCaseHeaders);
```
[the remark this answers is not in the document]
duncanista: Yes, I tested it – I also talked with the NodeJS tracer team, it seems that they don't do that there, so we are doing it here.

src/trace/context/extractor.spec.ts
duncanista: Not yet – sorry. The key thing to learn here is that, no matter what we test them with, we are mocking what the tracer returns. When the NodeJS tracer detects that it's on a testing environment, it will turn into a no-operation class. 

Either way, I'll add this suggestion and add some tests for this mock, since I think it's very useful at least to know what we are calling the tracer with. Maybe worth checking if we should test EVERYTHING, since that's a blackbox for us at the end of the day.

src/trace/context/extractors/http.spec.ts
duncanista: Yup!

src/trace/xray-service.ts
duncanista: Is that done in Xray? If so, then we should revise these values, either way, this is 4 year old code that I'm not aware which value is the right one!

src/trace/xray-service.ts
duncanista: Correct, but in typescript, you can de-structure arrays with fewer constant when an array contains more values to unpack. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

src/trace/xray-service.ts
purple4reina: Sounds good. Anything that's just a copy from the previous code, you're right we should just leave it how it was.

astuyve (commented):
A couple quick notes

src/trace/context/extractors/app-sync.ts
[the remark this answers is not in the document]
purple4reina: What if instead of using a `this.httpEventExtractor` you just use a regular extractor? Instead of passing an event object with a `headers` attr, you just pass the headers themselves?

src/trace/context/extractor.ts
```diff
@@ -0,0 +1,122 @@
...
+  async extract(event: any, context: Context): Promise<SpanContextWrapper | null> {
+    this.stepFunctionContextService = StepFunctionContextService.instance(event);
+
+    let spanContext: SpanContextWrapper | null = null;
+    if (this.config.traceExtractor) {
+      const customExtractor = new CustomTraceExtractor(this.config.traceExtractor);
+      spanContext = await customExtractor.extract(event, context);
+    }
+
+    if (spanContext === null) {
```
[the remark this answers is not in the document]
duncanista: Mmm, my current thought would be doing early returns, but this would imply to use the `this.addTraceContextToXray` method three times instead of one. That way, we avoid `null` checking, and we ensure early returns, wdyt?

src/trace/context/extractor.ts
```diff
@@ -0,0 +1,122 @@
...
+    if (spanContext !== null) {
+      this.addTraceContextToXray(spanContext);
+
+      return spanContext;
+    }
+
+    return this.xrayService.extract();
+  }
+
+  private getTraceEventExtractor(event: any): EventTraceExtractor | undefined {
```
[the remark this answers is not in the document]
duncanista: Do you mean declaring a type such as `DatadogTraceHeaders | SNSEvent | SQSEvent | KinesisEvent | ...`? and then plugging in it into this?

src/trace/context/extractors/app-sync.ts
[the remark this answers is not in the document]
duncanista: I'm not sure how AppSync events work, but I wonder if they could contain authorizer context – or headers with mixed casing, thus the original author using this approach – so, I agree with @astuyve on just moving `app-sync` into the http event extractor 🤔  Wdyt @purple4reina? Also, what do you mean by a "regular extractor"?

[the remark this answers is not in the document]
purple4reina: @duncanista okay apologies, I took a closer look at this and retract my comment about "regular extractor". I did not realize we were doing more with the event than just passing `event.headers` to the extractor.

I support moving `app-sync` into the http event extractor. 👍🏽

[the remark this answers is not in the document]
duncanista: Will do in another PR – to avoid making multiple non-w3c related changes here.

src/trace/context/extractor.spec.ts
duncanista: Logged a ticket to separately do this, since this is a non-block.

src/index.ts
```diff
@@ -20,8 +20,7 @@ import {
   setLogLevel,
 } from "./utils";
 import { getEnhancedMetricTags } from "./metrics/enhanced-metrics";
-
-export { TraceHeaders } from "./trace";
```
astuyve, line 24: We can't break this export for customers who may be using it.

astuyve (changes requested):

src/index.ts
```diff
@@ -20,8 +20,7 @@ import {
...
+import { DatadogTraceHeaders } from "trace/context/extractor";
```
astuyve, line 23: also somehow we broke the types here
image

duncanista: you have to provide relative path as `../../trace/my-file.ts`

astuyve: hehe yeah just a TODO note for myself

src/trace/context/extractor.ts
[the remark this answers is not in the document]
astuyve: Yeah I realize now you're copying this over anyway so let's leave that for later.

src/trace/context/extractor.ts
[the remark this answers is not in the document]
astuyve: Yes but same as above, let's leave it for now.

astuyve (approved):

purple4reina (approved):

PR #438, opened 2023-11-17
feat: use multiple processes to publish layers

astuyve:
### What does this PR do?
1. Revives some old code which spins off child processes (with `&`) and tracks the pids up to a batch size to make backfilling layers faster
2. *removes the backfill confirmation logic* and instead warns:
image
3. Removes the region arg, we already ask/confirm so if they don't pass it, we don't need to choose it:
image

### Motivation

I've never said "no" to a backfill, and I can't think of a condition where we would say no unless we inadvertently used a large number for the target version (eg 999 instead of 99). This script could simply be killed and started again with the lower number, as lambda will only let us increment versions upwards anyways.

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/scripts/publish_sandbox.sh
+++ b/scripts/publish_sandbox.sh
@@ -10,7 +10,7 @@ if [ -z "$VERSION" ]; then
 fi
 
 ./scripts/build_layers.sh
-VERSION=$VERSION REGIONS=us-east-1 ./scripts/publish_layers.sh
+VERSION=$VERSION ./scripts/publish_layers.sh
 
 # Automatically create PR against github.com/DataDog/documentation
 # If you'd like to test, please uncomment the below line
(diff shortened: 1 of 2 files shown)


scripts/publish_sandbox.sh
```diff
@@ -10,7 +10,7 @@ if [ -z "$VERSION" ]; then
 fi
 
 ./scripts/build_layers.sh
-VERSION=$VERSION REGIONS=sa-east-1 ./scripts/publish_layers.sh
+VERSION=$VERSION REGIONS=us-east-1 ./scripts/publish_layers.sh
```
purple4reina, line 13: should we keep this as `sa-east-1`?

astuyve: OH good question, this is intentional - our RC now runs in us-east-1 and production self monitoring runs in sa-east-1, so I think we want this change? Will ask Alex.

purple4reina (approved):

PR #456, opened 2023-12-20
feat: Fix FD leak in request

astuyve:
### What does this PR do?

Fully fixes the FD leak by discarding the `response` stream using `resume()`
We already discard the result of any GET or POST here, and always have (which we know from the types returning only a status code), so `resume()` is fine here.

I was under the impression that since we did not provide `response.on('data', cb)` that this clause wouldn't apply and the response stream would be ignored:

>If no [`'response'`](https://nodejs.org/api/http.html#event-response) handler is added, then the response will be entirely discarded. However, if a [`'response'`](https://nodejs.org/api/http.html#event-response) event handler is added, then the data from the response object **must** be consumed, either by calling `response.read()` whenever there is a `'readable'` event, or by adding a `'data'` handler, or by calling the `.resume()` method. Until the data is consumed, the `'end'` event will not fire. Also, until the data is read it will consume memory that can eventually lead to a 'process out of memory' error.

but of course passing a callback function _counts_ as a `handler` so in this case the **must** clause counts and the underlying readable stream was never consumed, causing the FD leak.

This code has been present for many years but only became an issue in Node20. For that, I'm not exactly clear. But I can confirm this PR fixes the issue.

### Motivation

You can see the fix in this [notebook](https://ddserverless.datadoghq.com/notebook/7135653/aj-node20-file-descriptor-leak)

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/utils/request.ts
+++ b/src/utils/request.ts
@@ -55,7 +55,7 @@ function sendRequest(url: URL, options: RequestOptions, buffer?: Buffer): Promis
 
       // https://nodejs.org/api/http.html#class-httpclientrequest
       // "Until the data is consumed, the 'end' event will not fire. Also, until the data is read it will consume memory that can eventually lead to a 'process out of memory' error"
-      response.resume()
+      response.resume();
 
       if (statusCode === undefined || statusCode < 200 || statusCode > 299) {
         return resolve({
--- a/src/utils/request.ts
+++ b/src/utils/request.ts
@@ -40,12 +40,20 @@ export function get(url: URL, options?: Partial<RequestOptions>): Promise<Reques
   return sendRequest(url, requestOptions);
 }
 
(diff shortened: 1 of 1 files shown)


src/utils/request.ts
```diff
@@ -61,7 +71,7 @@ function sendRequest(url: URL, options: RequestOptions, buffer?: Buffer): Promis
       });
     });
 
-    request.on("error", (error) => {
+    request.once("error", (error) => {
```
astuyve, line 74: https://nodejs.org/api/events.html#emitteronceeventname-listener

this just guarantees we don't leave an emitter dangling

joeyzhao2018 (approved):
Wow

PR #460, opened 2024-01-10
fix: block forbidden chars from metric names

dotkas:
### What does this PR do?

Block calling metric functions with names that will cause `datadog-agent` to fail.

### Motivation

I should probably have known better, but it took me a merged PR all the way to prod and an error message in DataDog Serverless before I realized my code didn't work, and a clone of `datadog-agent` Go code to realize what I'd messed up. So I figured this could help other people.

### Testing Guidelines

I added a new Jest case.

### Additional Notes

I guess this is technically a breaking change. But if anybody would be affected by it we'd really be doing them a favor here as they would still have errors in their logs, only further down the pipeline. 🤷🏻 

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -15,12 +15,13 @@ import {
   Logger,
   LogLevel,
   promisifiedHandler,
-  setSandboxInit,
   setLogger,
   setLogLevel,
+  setSandboxInit,
 } from "./utils";
 import { getEnhancedMetricTags } from "./metrics/enhanced-metrics";
 import { DatadogTraceHeaders } from "./trace/context/extractor";
+import { MalformedMetricName } from "./runtime/errors";
 
 // Backwards-compatible export, TODO deprecate in next major
 export { DatadogTraceHeaders as TraceHeaders } from "./trace/context/extractor";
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -255,6 +260,14 @@ export function extractArgs<TEvent>(isResponseStreamFunction: boolean, ...args:
  * @param tags The tags associated with the metric. Should be of the format "tag:value".
  */
 export function sendDistributionMetricWithDate(name: string, value: number, metricTime: Date, ...tags: string[]) {
+  if (FORBIDDEN_METRIC_NAME_CHARS.some((c) => name.includes(c))) {
+    throw new MalformedMetricName(
```
astuyve, line 264: This would cause the customer's lambda function to crash, which I don't think we'd want to do at any point

astuyve (changes requested):
Hi, thanks so much for taking the time to write this PR up! I'm sorry that you ran into this metric naming issue.
This is a good PR, technically well written, conforms with our conventions, and includes good tests. I appreciate your hard work!

I'm not sure if this something we'd merge at this time, but I'm open to other feedback from additional users and we can revisit this conclusion in the future.

My reasoning is that we do document the format of metric names both implicitly in our [lambda examples](https://docs.datadoghq.com/serverless/aws_lambda/metrics/?code-lang=python#submit-custom-metrics-directly-from-a-lambda-function) and explicitly [in the datadog metric documentation](https://docs.datadoghq.com/developers/guide/what-best-practices-are-recommended-for-naming-metrics-and-tags/#rules-and-best-practices-for-naming-metrics).

If we did choose to validate the format in the Lambda function itself, we'd prefer to implement it in the Datadog Lambda extension, as this would benefit all users across all runtimes (not just nodejs). That also allows us to change the format in one place if that is ever needed in the future.

Finally this implementation does throw an unhandled exception when an invalid metric name is sent, which we wouldn't want to do as it would crash the running Lambda function.

We can leave this PR in an open state for now and see if other users would like this feature, but I don't intend to merge it at this time.

Thanks again!

Issue #482, opened 2024-02-02
"Cannot read properties of undefined (reading 'toString')" when calling getTraceHeaders()


midu-git:
## Expected Behavior
Calling getTraceHeaders() should not error out and provide trace headers correctly.

## Actual Behavior
We are using this guy from a lambda called on the s3 middleware. At least in some situations, we encounter this error

```
{
    errorType: "TypeError",
    errorMessage: "Cannot read properties of undefined (reading 'toString')",
    stack: [
        "TypeError: Cannot read properties of undefined (reading 'toString')",
        "    at TraceContextService.get (/opt/nodejs/node_modules/datadog-lambda-js/trace/trace-context-service.js:87:93)",
        "    at TraceListener.get (/opt/nodejs/node_modules/datadog-lambda-js/trace/listener.js:76:40)",
        "    at getTraceHeaders (/opt/nodejs/node_modules/datadog-lambda-js/index.js:363:33)",
        "    at injectDatadogHeadersIntoS3 (/shared/src/utils/tracing-utils.ts:71:28)",
        "    at next (/shared/src/utils/s3-utils.ts:219:17)",
        "    at <anonymous> (/node_modules/@smithy/middleware-apply-body-checksum/dist-cjs/applyMd5BodyChecksumMiddleware.js:29:12)",
        "    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)",
    ],
};
```

We have identified the error to [this code](https://github.com/DataDog/datadog-lambda-js/blob/main/src/trace/span-context-wrapper.ts#L21)
```
public sampleMode(): number {
  return this.spanContext._sampling?.priority;
}
```
Basically, you allow a number to be returned, but it should be `number | undefined`.

## Steps to Reproduce the Problem

  1.
  1.
  1.

## Specifications

  - Datadog Lambda Layer version: layer:Datadog-Node18-x:105
  - Node version: 18

## Stacktrace

  ```
        "TypeError: Cannot read properties of undefined (reading 'toString')",
        "    at TraceContextService.get (/opt/nodejs/node_modules/datadog-lambda-js/trace/trace-context-service.js:87:93)",
        "    at TraceListener.get (/opt/nodejs/node_modules/datadog-lambda-js/trace/listener.js:76:40)",
        "    at getTraceHeaders (/opt/nodejs/node_modules/datadog-lambda-js/index.js:363:33)",
[...]

PR #483, opened 2024-02-07
fix: reading sampling priority will error when not available

duncanista:
### What does this PR do?

Fixes a bug in which reading sampling priority will cause an error, because it was undefined. Previously, before adding W3C, we'd just set always `AUTO_KEEP`, but now we decided on getting whatever the trace context had. 

Now, if not available, we'll go back to using `AUTO_KEEP`, while respecting whichever was there before if set.

### Motivation

#482 

### Testing Guidelines

Added unit tests.

### Additional Notes

Introduced in: https://github.com/DataDog/datadog-lambda-js/commit/f05c6003a4f7bab82e46e741ac8d0c98d89efe5b#diff-8367049249917f97a7d4b0d7aa3355ac631ce80bd38d35f0f04c9288c603b76a

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/trace/span-context-wrapper.spec.ts
+++ b/src/trace/span-context-wrapper.spec.ts
@@ -23,25 +23,31 @@ describe("SpanContextWrapper", () => {
 
   describe("sampleMode", () => {
     it("should return AUTO_KEEP when sampling priority is not available in spanContext", () => {
-      const spanContext = new SpanContextWrapper({
-        toSpanId: () => "1234",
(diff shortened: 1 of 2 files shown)


astuyve (approved):
Good use of null coalescence here, `0 ||` would be falsey and could cause us to keep everything and break sampling (if priority is an int)

PR #513, opened 2024-03-15
Favor parameterized routes for resource name.

purple4reina:
### What does this PR do?

Set the resource name to the parameterized route key when using api gateway v2.

### Motivation

A metric is created from each span made by the tracers. The metric name comes from the operation name. The resource name gets added as a tag. Therefore unbound values for resource name can cause a large increase in the number of metrics on an account.

### Testing Guidelines

To test this, I created two apps and pulled the event payloads for each to use in the unit tests.

```yaml
functions:
  v1:
    handler: handler.handler
    events:
      - http:
          method: get
          path: /user/{id}
  v2:
    handler: handler.handler
    events:
      - httpApi:
          method: get
          path: /user/{id}
```

### Additional Notes

This feature already worked for api gatewawy v1.

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/trace/span-inferrer.ts
+++ b/src/trace/span-inferrer.ts
@@ -525,7 +525,7 @@ export class SpanInferrer {
   }
 
   getResourcePath(event: any): string {
-    const routeKey = event.requestContext.routeKey;
+    const routeKey = event?.requestContext?.routeKey;
     if (routeKey && routeKey.includes("{")) {
       // this is a parameterized route
       try {
--- a/src/trace/span-inferrer.ts
+++ b/src/trace/span-inferrer.ts
@@ -94,15 +94,7 @@ export class SpanInferrer {
     const options: SpanOptions = {};
     const domain = event.requestContext.domainName || "";
     const path = event.rawPath || event.requestContext.path || event.requestContext.routeKey;
-    var resourcePath = event.rawPath || event.requestContext.resourcePath || event.requestContext.routeKey;
-    if (event.requestContext.routeKey && event.requestContext.routeKey.includes("{")) {
-        // this is a parameterized route
-        try {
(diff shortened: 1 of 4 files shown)


src/trace/span-inferrer.ts
```diff
@@ -94,7 +94,15 @@ export class SpanInferrer {
     const options: SpanOptions = {};
     const domain = event.requestContext.domainName || "";
     const path = event.rawPath || event.requestContext.path || event.requestContext.routeKey;
-    const resourcePath = event.rawPath || event.requestContext.resourcePath || event.requestContext.routeKey;
+    var resourcePath = event.rawPath || event.requestContext.resourcePath || event.requestContext.routeKey;
```
astuyve, line 97: I'd suggest refactoring this line + 98 into a new method with your new conditional as the first path, and use early returns as control flow.

My concern would be that we shadow/overwrite `resourcePath` conditionally here in 101, and that if we add new logic here then someone needs to modify the logic on 97 they may not realize that their change would be later overwritten if it's a parameterized route.

Is that helpful?

src/trace/span-inferrer.ts
```diff
@@ -523,4 +523,17 @@ export class SpanInferrer {
     };
     return new SpanWrapper(this.traceWrapper.startSpan("aws.eventbridge", options), spanWrapperOptions);
   }
+
+  getResourcePath(event: any): string {
+    const routeKey = event.requestContext.routeKey;
```
astuyve, line 528: Optional chaining so this can be called safely given the `any` type. Your checks on 529 are enough otherwise.

```suggestion
    const routeKey = event?.requestContext?.routeKey;
```

astuyve (approved):
One note, everything else looks good

PR #516, opened 2024-03-18
feat: Remove extension hello check as it's redundant

astuyve:
### What does this PR do?
Checking the file presence is enough, start order is enforced by the lambda runtime.

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/metrics/extension.spec.ts
+++ b/src/metrics/extension.spec.ts
@@ -11,19 +11,8 @@ describe("isAgentRunning", () => {
     mock({
       "/opt/extensions/datadog-agent": Buffer.from([0]),
     });
-    const scope = nock(AGENT_URL).get("/lambda/hello").reply(200);
     const ran = await isAgentRunning();
-    expect(scope.isDone()).toBeTruthy();
     expect(ran).toBeTruthy();
-  });
-  it("returns false when agent doesn't respond", async () => {
-    mock({
-      "/opt/extensions/datadog-agent": Buffer.from([0]),
-    });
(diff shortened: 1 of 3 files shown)


duncanista (approved):
LGTM

purple4reina:
So. I know the extension specifically uses the call to `Hello` to determine if there's a tracing layer present. See https://github.com/DataDog/datadog-agent/blob/11e8ea8b5455530ba8afa8ef1831d179c5523479/pkg/serverless/daemon/routes.go#L29

astuyve:
We do, but the library never calls `start-invocation`, so I assume the call to `enrichInferredSpan()` never happens or happens but the result is never used. [source](https://github.com/DataDog/datadog-agent/blob/11e8ea8b5455530ba8afa8ef1831d179c5523479/pkg/serverless/invocationlifecycle/init.go#L28).

We can see with this build, there are no duplicate spans. And we have made this change in Python already.

purple4reina:
> We do, but the library never calls `start-invocation`, so I assume the call to `enrichInferredSpan()` never happens or happens but the result is never used. [source](https://github.com/DataDog/datadog-agent/blob/11e8ea8b5455530ba8afa8ef1831d179c5523479/pkg/serverless/invocationlifecycle/init.go#L28).
> 
> We can see with this build, there are no duplicate spans. And we have made this change in Python already.

Ah, yes I see now. Thanks for the explanation. We can probably just go ahead and delete that code from the extension then yeah?

Issue #58, opened 2020-03-09
Allow set custom timestamp for metric datapoint


PatrykMilewski:
## Expected Behavior

When I'm sending datapoint with `sendDistributionMetric` function, I would like to specify my custom timestamp.

## Actual Behavior

Right now it automatically uses the actual time for timestamp generation.

## Specifications

  - Datadog Lambda Layer version: none, datadog-lambda-layer-js:2.17.0
  - Node version: 12.x

agocs:
👋 Hi, I've added this to a list of feature requests. Custom timestamps in this library make sense. We already allow them in Ruby, Python, and Go.

PatrykMilewski:
@agocs Could you specify the header of the function, that would have this feature? I'm not sure how you are planning you API for this package, but if you guys would give me some hints, then I could implement it myself and create PR.

agocs:
Hey @PatrykMilewski , I definitely appreciate the offer. I think the new function would take the signature

```
export function sendDistributionMetricWithDate(name: string, value: number, date: Date, ...tags: string[])
```

It would, unfortunately, require refactoring [`listener.sendDistrobutionMetric`](https://github.com/DataDog/datadog-lambda-layer-js/blob/master/src/metrics/listener.ts#L86) and a few other functions to accept the custom date.

vishnuvisnu:
`public sendDistributionMetricWithDate(name: string, value: number, date: Date, ...tags: string[]) {
    if (this.config.logForwarding) {
      writeMetricToStdout(name, value, tags);
      return;
    }
    const dist = new Distribution(name, [{ timestamp: date, value }], ...tags);

    if (this.currentProcessor !== undefined) {
      this.currentProcessor.then((processor) => {
        processor.addMetric(dist);
      });
    } else {
      logError("can't send metrics, datadog lambda handler not set up.");
    }
}

public sendDistributionMetric(name: string, value: number, ...tags: string[]) {
  sendDistributionMetricWithDate(name, value, new Date(), tags);
}
`
These changes to MetricListener in listener.ts along with new function sendDistributionMetricWithDate in index.ts should work, right?

vishnuvisnu:
@agocs This is the PR https://github.com/DataDog/datadog-lambda-layer-js/pull/63 I have created.

vishnuvisnu:
[...]

PR #519, opened 2024-03-18
chore: lazy load internal request module

duncanista:
### What does this PR do?

Reduces memory allocation for this package by lazy loading internal `request.ts` methods.

Before:
Screenshot 2024-03-20 at 3 19 36 PM

After:
Screenshot 2024-03-20 at 3 17 40 PM

### Motivation

It was possible to reduce the memory allocations + performance.

### Testing Guidelines

- Integration tests remained the same.
- Unit tests passed as normal
- Memory allocation was reduced.

### Additional Notes

[SVLS-4628](https://datadoghq.atlassian.net/browse/SVLS-4628)

Allocation time can vary, but now we are sure that the lazy load is in effect.

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Misc (docs, refactoring, dependency upgrade, etc.)

### Check all that apply

- [ ] This PR's description is comprehensive
- [ ] This PR contains breaking changes that are documented in the description
- [ ] This PR introduces new APIs or parameters that are documented and unlikely to change in the foreseeable future
- [ ] This PR impacts documentation, and it has been updated (or a ticket has been logged)
- [ ] This PR's changes are covered by the automated tests
- [ ] This PR collects user input/sensitive content into Datadog
- [ ] This PR passes the integration tests (ask a Datadog member to run the tests)

[SVLS-4628]: https://datadoghq.atlassian.net/browse/SVLS-4628?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #522, opened 2024-03-20
[SVLS-4422] Support metrics with timestamps when the Extension is running

DylanLovesCoffee:
### What does this PR do?

Allows `sendDistributionMetricWithDate` to submit a metric with a timestamp when the Extension is running. Neither hot-shots nor Dogstatsd supports historical distribution metrics (more on the latter [here](https://github.com/Datadog/architecture/blob/master/rfcs/dogstatsd-no-aggregation-pipeline/rfc.md) and [here](https://datadoghq.atlassian.net/browse/FRMETS-1843)).

The ability to send metrics with timestamps already existed but did not work when using the Extension.

### Motivation

* [SVLS-4422](https://datadoghq.atlassian.net/browse/SVLS-4422)
* Related to: 
  * **TODO**: https://github.com/DataDog/datadog-lambda-python/issues/398 
  * https://github.com/DataDog/datadog-lambda-js/issues/58 

### Testing Guidelines

Unit test + manual tests. Example of one invocation using the following:
```js
const MS_PER_MINUTE = 60000;
const timestamp = new Date(Date.now() - MS_PER_MINUTE);
sendDistributionMetricWithDate("dylan.test.past", 1, timestamp);
sendDistributionMetric("dylan.test.present", 1); // tracer.dogstatsd.distribution("dylan.test.present", 1) also works
```
Screenshot 2024-03-20 at 12 34 36 PM

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

[SVLS-4422]: https://datadoghq.atlassian.net/browse/SVLS-4422?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/metrics/listener.ts
+++ b/src/metrics/listener.ts
@@ -153,7 +153,7 @@ export class MetricsListener {
           processor.addMetric(dist);
         });
         return;
-      };
+      }
 
       this.statsDClient?.distribution(name, value, undefined, tags);
       return;
--- a/src/metrics/listener.ts
+++ b/src/metrics/listener.ts
@@ -131,7 +131,7 @@ export class MetricsListener {
   public sendDistributionMetricWithDate(
     name: string,
     value: number,
-    metricTime: Date,
+    metricTime: Date, // TODO: Next breaking change to update to optional or 'Date | undefined'?
     forceAsync: boolean,
     ...tags: string[]
   ) {
--- a/src/metrics/listener.ts
+++ b/src/metrics/listener.ts
@@ -140,7 +140,21 @@ export class MetricsListener {
     forceAsync: boolean,
     ...tags: string[]
   ) {
+    const dist = new Distribution(name, [{ timestamp: metricTime, value }], ...tags);
+
     if (this.isExtensionRunning) {
+      // Dogstatsd doesn't support distribution metrics with timestamps so we must use the API
+      const isMetricTimeValid = Date.parse(metricTime.toString()) > 0;
+      if (isMetricTimeValid) {
+        if (this.currentProcessor === undefined) {
+          this.currentProcessor = this.createProcessor(this.config, this.apiKey);
+        }
+        this.currentProcessor.then((processor) => {
+          processor.addMetric(dist);
+        });
+        return;
+      };
+
       this.statsDClient?.distribution(name, value, undefined, tags);
       return;
     }
@@ -148,8 +162,6 @@ export class MetricsListener {
       writeMetricToStdout(name, value, metricTime, tags);
       return;
     }
-
-    const dist = new Distribution(name, [{ timestamp: metricTime, value }], ...tags);
 
     if (!this.apiKey) {
       const errorMessage = "api key not configured, see https://dtdg.co/sls-node-metrics";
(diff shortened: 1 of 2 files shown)


duncanista:
Do not merge before #519

DylanLovesCoffee:
> Do not merge before #519

should I just change the base to `jordan.gonzalez/lazy-load-fallback-processor`?

src/metrics/listener.ts
```diff
@@ -167,7 +174,9 @@ export class MetricsListener {
   }
 
   public sendDistributionMetric(name: string, value: number, forceAsync: boolean, ...tags: string[]) {
-    this.sendDistributionMetricWithDate(name, value, new Date(Date.now()), forceAsync, ...tags);
+    // The Extension doesn't support distribution metrics with timestamps. Use sendDistributionMetricWithDate instead.
+    const metricTime = this.isExtensionRunning ? new Date(0) : new Date(Date.now());
```
duncanista, line 178: What's the purpose of setting a metric time if the extension is still gonna ignore it? (as understood by the purpose of the PR)

src/metrics/listener.ts
```diff
@@ -135,17 +135,24 @@ export class MetricsListener {
...
+
     if (this.isExtensionRunning) {
-      this.statsDClient?.distribution(name, value, undefined, tags);
-      return;
+      const isMetricTimeValid = Date.parse(metricTime.toString()) > 0;
+      if (isMetricTimeValid) {
+        // Only create the processor to submit metrics to the API when a user provides a valid timestamp as
+        // Dogstatsd does not support timestamps for distributions.
+        this.currentProcessor = this.createProcessor(this.config, this.apiKey);
+      } else {
```
duncanista, line 146: Let's try to avoid having `if-else` blocks. Also, wouldn't this need a return? This seems like it would process the metric if the extension is running but if `this.config.logForwarding` is set to `true`, it will both process it and then write it to `stdout`.

src/metrics/listener.ts
DylanLovesCoffee: I wasn't too sure if changing the type for this function's parameter would introduce a breaking change to users or not, so I figured I'd rather just not change the type of the `metricTime` param from `Date` -> `Date | undefined`. But if it's not going to change much then I can just update the type to `Date | undefined` and pass undefined here

src/metrics/listener.ts
DylanLovesCoffee: The `if` here just creates the processor since it's typically never needed when using the extension, except for this special case, otherwise always send to statsd. Later on the metric is added to the processor at the end of the function (unchanged). So if `logForwarding` is true and the extension is running, then it will return on L153 and not process

duncanista: You are right, thanks for explaining!

src/metrics/listener.ts
duncanista: I think the workaround makes sense. Changing the API would definitely make a breaking change, so that's a no in the mean time. Can you leave a note in the form of `// TODO: Next breaking change, change API for metric`.

DylanLovesCoffee: yessir 👍

duncanista (approved):
LGTM – thanks for fixing this!

Issue #515, opened 2024-03-18
datadog: handler not initialized


naveen700:
## Expected Behavior
Metric should have been sent to datadog in lambda instead getting error of handler not initialized.

#369 
## Actual Behavior

I am using sendDistributionMetric but while checking the lambda logs I am getting an error as datadog: handler not initialized.
## Steps to Reproduce the Problem

  1. Using dd-trace and initializing it using init() method
  2. Then wrapping the handler in trace.
  3. And using method sendDistributionMetric  to send the metric.

## Specifications

  - Datadog Lambda Layer version: 7.96.0
  - dd trace: 5.6.0
  - Node version: Node 18.x

## Stacktrace

2024-03-18T05:10:34.021Z	79214b2e-b340-469e-ac86-ec73b4e0ee51	ERROR	[dd.trace_id=5898828417058474788 dd.span_id=1717120647592796076] 
{
    "status": "error",
    "message": "datadog:handler not initialized"
}
@astuyve I have seen your comment and added the logs as sent and send metric as you mentioned in ticket #369 
![image](https://github.com/DataDog/datadog-lambda-js/assets/23037155/831878e3-c3c3-43e8-b8d7-5e4e1b4104d0)

astuyve:
Hi @naveen700 - thanks for reaching out!

You shouldn't need to manually initialize DDtrace at all, this library handles that [here](https://github.com/DataDog/datadog-lambda-js/blob/cb2656d3275a091e9e18479ab146fd04c3861381/src/runtime/module_importer.js#L15). You also shouldn't need to manually wrap your handler, but without seeing that code it's hard to for me to help more.

This error (like the linked issue) is occurring because `sendDistributionMetric` is being called _before_ the `Datadog()` method you'd use to wrap your handler. Typically this is because a user is attempting to send a metric during the initialization phase of the Lambda function.

We'll work towards buffering these metrics so this error doesn't occur, but if you're still having trouble I'd ask that you file a support ticket so that we can see your configuration and handler code and offer specific advice. You can attach this ticket to speed up the process.

Support can be reached at `support@datadoghq.com`, or using the Live Chat feature in the application.

Thanks!

naveen700:
Hi @astuyve,

[...]

PR #524, opened 2024-03-22
feat: support sending metrics before handler initialization

duncanista:
### What does this PR do?

Allows customer to use the API `sendDistributionMetric` and `sendDistributionMetricWithDate` outside of the handler.

### Motivation

#515 

### Testing Guidelines

- Added unit tests
- Added integration test
- Tested manually

### Additional Notes

[SVLS-4633](https://datadoghq.atlassian.net/browse/SVLS-4633)

### Types of Changes

- [x] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

[SVLS-4633]: https://datadoghq.atlassian.net/browse/SVLS-4633?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -279,8 +279,12 @@ export function sendDistributionMetric(name: string, value: number, ...tags: str
 }
 
 function sendQueueMetrics(listener: MetricsListener) {
+  // Reverse the queue to send metrics in order.
+  // This is necessary because the "queue" is a stack,
+  // and we want to send metrics in the order they were added.
+  _metricsQueue.reverse();
   while (_metricsQueue.length > 0) {
-    const metric = _metricsQueue.shift()!; // This will always exist.
+    const metric = _metricsQueue.pop()!; // This will always exist.
     const { name, value, metricTime, tags } = metric;
     if (metricTime !== undefined) {
       listener.sendDistributionMetricWithDate(name, value, metricTime, false, ...tags);
--- a/src/index.ts
+++ b/src/index.ts
@@ -12,7 +12,6 @@ import { TraceConfig, TraceListener } from "./trace";
 import { subscribeToDC } from "./runtime";
 import {
   logDebug,
-  logError,
   Logger,
   LogLevel,
   promisifiedHandler,
--- a/src/index.ts
+++ b/src/index.ts
@@ -284,7 +284,7 @@ function sendQueueMetrics(listener: MetricsListener) {
     const { name, value, metricTime, tags } = metric;
     if (metricTime !== undefined) {
       listener.sendDistributionMetricWithDate(name, value, metricTime, false, ...tags);
-      return;
+      continue;
     }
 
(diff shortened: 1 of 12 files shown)


src/index.ts
```diff
@@ -275,8 +272,22 @@ export function sendDistributionMetric(name: string, value: number, ...tags: str
...
+  _metricsQueue.push({ name, value, tags });
+}
+
+function sendQueueMetrics(listener: MetricsListener) {
+  while (_metricsQueue.length > 0) {
+    const metric = _metricsQueue.shift()!; // This will always exist.
+    const { name, value, metricTime, tags } = metric;
+    if (metricTime !== undefined) {
+      listener.sendDistributionMetricWithDate(name, value, metricTime, false, ...tags);
+      return;
```
joeyzhao2018, line 287: should this be `continue`?

src/index.ts
```diff
@@ -275,8 +272,22 @@ export function sendDistributionMetric(name: string, value: number, ...tags: str
...
-    logError("handler not initialized");
+    return;
+  }
+
+  _metricsQueue.push({ name, value, tags });
+}
+
+function sendQueueMetrics(listener: MetricsListener) {
+  while (_metricsQueue.length > 0) {
+    const metric = _metricsQueue.shift()!; // This will always exist.
```
joeyzhao2018, line 283: two question on the `shift()` part:
1. i think it's O(n) time complex. why don't we just use an index to loop.
2. i assume because we also want to clean up the queue. this makes me wonder if we could run into some corner cases such as timeouts and results in a dangling queue?

src/index.ts
duncanista: Yup, you are right, thanks – great catch!

src/index.ts
duncanista: Ahhh, you are right, this is definitely slow. 

I might either:
1. Use the `pop` method since it's `O(1)`, but metrics will arrive in a LIFO order, ending in a `O(n)`.
2. Use the previous approach, but reverse before sending data, instead of being `O(n ** 2)`, it would be at most `O(2n)` => `O(n)`, WDYT?

> i assume because we also want to clean up the queue. this makes me wonder if we could run into some corner cases such as timeouts and results in a dangling queue?
Not worried about timeouts, since this operation should be really fast, even with a thousand metrics.
On dangling queues, this shouldn't be possible, since we clean it on every flush. If there's no flush, then once the queue is full, it will be reset. WDYT?

joeyzhao2018 (approved):
LGTM

src/index.ts
duncanista: Updated and use the second strategy as per our conversation!

PR #534, opened 2024-05-02
breaking: Use sha256 to hash lambda traceId that are triggered by Step Functions  and set _dd.p.tid

kimi-p:
### What does this PR do?
- using the most significant 128 bits of sha256 (instead of current md5) to hash trace_id and the first 64 for span_id.
- the 1st and 65th bit should be `0`.
- manually take:
  - the first 64 bits from the sha256 hash for `_dd.p.tid` tag
  - the 2nd 64 bits from the sha256 hash for `traceId`
  - if it were for spanId, use sha256 hash for `parentId` (which is step function task's spanId)
- `ts-md5` package is remove.
- @node/type in DevDependencies needs to be upgraded by running `yarn add @types/node -D` as suggested in this the issue https://github.com/microsoft/TypeScript/issues/51567
- note that for the root span it has span_id in number but trace_id in hex
- [relevant PR for 64 bits using md5](https://github.com/DataDog/datadog-lambda-js/pull/331)
```
{
    "trace": {
        "root_id": "3124668925291824462",
        "spans": {
            "3124668925291824462": {
                "trace_id": "41919d29ba4149cd40fcae25a0e4e323",
                "span_id": "3124668925291824462",
                "parent_id": "0",
                "start": 1716393574.422,
                "end": 1716393576.138,
                "duration": 1.716,
                "type": "stepfunctions",
                "service": "kimi-test-nodejs-linking",
                "name": "aws.stepfunctions",
                "resource": "kimi-test-nodejs-linking",
                "resource_hash": "c76d3199bc4b077",
                "host_id": 3884291,
                "hostname": "none",
                "env": "dev",
                "host_groups": [
                    "env:dev"
                ],
...
```

### Motivation
To support 128 bits trace IDs and to avoid showing up in security vulnerability scans, we are upgrading the hashing method (md5) to sha256.

### Testing Guidelines
- [trace link](https://dd.datad0g.com/functions?cloud=aws&entity_view=step_functions&fromUser=false&graphType=flamegraph&panel_end=1716393588294&panel_paused=false&panel_start=1716389988294&shouldShowLegend=true&sort=time&sp=%5B%7B%22p%22%3A%7B%22entityId%22%3A%22aws-step-functions%2Barn%3Aaws%3Astates%3Asa-east-1%3A425362996713%3Astatemachine%3Akimi-test-nodejs-linking%22%7D%2C%22i%22%3A%22step-function-panel%22%7D%2C%7B%22p%22%3A%7B%22traceID%22%3A%2241919d29ba4149cd40fcae25a0e4e323%22%2C%22selectedSpanID%22%3A%223124668925291824462%22%7D%2C%22i%22%3A%22trace-panel%22%7D%5D&spanID=3124668925291824462&text_search=linking&traceID=41919d29ba4149cd40fcae25a0e4e323&view=spans&start=1716389688357&end=1716393288357&paused=false)
- example traceId `41919d29ba4149cd40fcae25a0e4e323` is a hash, not a number anymore.
- verified that the parent_id of the `aws.lambda` span matches the `aws.stepfunctions.lambda` task span.

image

### Additional Notes
This is a breaking change that will affect all current Step Functions public beta users. And what will break is the Step Functions trace and Lambda traces linking. 

Since this has to be done at some point, we want to put this change out before GA. Serverless-Integrations team will work with Sumedha and have some sort of announcement to let customers know the exact cut off time so that they can deploy their Lambda around the same time.

[Referencing PR for implementing MD5](https://github.com/DataDog/datadog-lambda-js/pull/331)

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [x] Breaking change

### Check all that apply

- [x] This PR contains breaking changes that are documented in the description

### DIFF
--- a/src/trace/step-function-service.spec.ts
+++ b/src/trace/step-function-service.spec.ts
@@ -194,8 +194,8 @@ describe("StepFunctionContextService", () => {
 
       expect(spanContext).not.toBeNull();
 
-      expect(spanContext?.toTraceId()).toBe("947965466153612645");
-      expect(spanContext?.toSpanId()).toBe("4602916161841036335");
+      expect(spanContext?.toTraceId()).toBe("3661440683");
+      expect(spanContext?.toSpanId()).toBe("2846425757");
       expect(spanContext?.sampleMode()).toBe("1");
       expect(spanContext?.source).toBe("event");
     });
@@ -215,7 +215,7 @@ describe("StepFunctionContextService", () => {
     it("returns the same hash number generated in `logs backend` for a random string", () => {
       const instance = StepFunctionContextService.instance();
       const hash = instance["deterministicMd5HashToBigIntString"]("some_testing_random_string");
-      expect(hash).toEqual("2251275791555400689");
+      expect(hash).toEqual("80506605202309154694697844088692857990");
     });
 
     it("returns the same hash number generated in `logs backend` for execution id # state name # entered time", () => {
@@ -223,7 +223,7 @@ describe("StepFunctionContextService", () => {
       const hash = instance["deterministicMd5HashToBigIntString"](
         "arn:aws:states:sa-east-1:601427271234:express:DatadogStateMachine:acaf1a67-336a-e854-1599-2a627eb2dd8a:c8baf081-31f1-464d-971f-70cb17d01111#step-one#2022-12-08T21:08:19.224Z",
       );
-      expect(hash).toEqual("8034507082463708833");
+      expect(hash).toEqual("80072419077927731656239868244106251139");
     });
   });
 
@@ -232,22 +232,22 @@ describe("StepFunctionContextService", () => {
       [
         "a random string",
         "some_testing_random_string",
-        "0001111100111110001000110110011110010111000110001001001111110001",
+        "00111100100100010000001000010111010011000111001011011001111000000110011101111001100001011100111110110001011111001101110010000110",
       ],
       [
         "an execution id",
         "arn:aws:states:sa-east-1:601427271234:express:DatadogStateMachine:acaf1a67-336a-e854-1599-2a627eb2dd8a:c8baf081-31f1-464d-971f-70cb17d041f4",
-        "0010010000101100100000101011111101111100110110001110111100111101",
+        "01000101001100100100101000010110011101001110110101001100100001000100010111011110010011011100010100101011110110011010110001111110",
       ],
       [
         "another execution id",
         "arn:aws:states:sa-east-1:601427271234:express:DatadogStateMachine:acaf1a67-336a-e854-1599-2a627eb2dd8a:c8baf081-31f1-464d-971f-70cb17d01111",
-        "0010001100110000011011011111010000100111100000110000100100101010",
+        "00101111100011001000100001010011001100000000000101110111001010110100110111010111011001101001111001110001011111000111010010101001",
       ],
       [
         "execution id # state name # entered time",
         "arn:aws:states:sa-east-1:601427271234:express:DatadogStateMachine:acaf1a67-336a-e854-1599-2a627eb2dd8a:c8baf081-31f1-464d-971f-70cb17d01111#step-one#2022-12-08T21:08:19.224Z",
-        "0110111110000000010011011001111101110011100111000000011010100001",
+        "00111100001111010110001100001111111100111110101100000000001000010011110011111110111010000100011111010100111110011000101110000011",
       ],
     ])("returns the same hash number generated in `logs backend` for %s", (_, str, expected) => {
       const instance = StepFunctionContextService.instance();
@@ -276,27 +276,16 @@ describe("StepFunctionContextService", () => {
     });
   });
 
-  describe("hexToBinary", () => {
+  describe("numberToBinaryString", () => {
     const instance = StepFunctionContextService.instance();
     it.each([
-      ["0", "0000"],
-      ["1", "0001"],
-      ["2", "0010"],
-      ["3", "0011"],
-      ["4", "0100"],
-      ["5", "0101"],
-      ["6", "0110"],
-      ["7", "0111"],
-      ["8", "1000"],
-      ["9", "1001"],
-      ["a", "1010"],
-      ["b", "1011"],
-      ["c", "1100"],
-      ["d", "1101"],
-      ["e", "1110"],
-      ["f", "1111"],
+      [0, "00000000"],
+      [1, "00000001"],
+      [2, "00000010"],
+      [3, "00000011"],
+      [4, "00000100"],
     ])("returns the right binary number for %s => %s", (hex, expected) => {
-      const binary = instance["hexToBinary"](hex);
+      const binary = instance["numberToBinaryString"](hex);
       expect(binary).toBe(expected);
     });
   });
--- a/src/trace/step-function-service.spec.ts
+++ b/src/trace/step-function-service.spec.ts
@@ -1,4 +1,4 @@
-import { _128_BITS, _64_BITS, StepFunctionContextService } from "./step-function-service";
+import { PARENT_ID, StepFunctionContextService } from "./step-function-service";
 
 describe("StepFunctionContextService", () => {
   const stepFunctionEvent = {
@@ -194,7 +194,7 @@ describe("StepFunctionContextService", () => {
 
       expect(spanContext).not.toBeNull();
 
-      expect(spanContext?.toTraceId()).toBe("3661440683");
+      expect(spanContext?.toTraceId()).toBe("1139193989631387307");
       expect(spanContext?.toSpanId()).toBe("5892738536804826142");
       expect(spanContext?.sampleMode()).toBe("1");
       expect(spanContext?.source).toBe("event");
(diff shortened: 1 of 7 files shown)


src/trace/step-function-service.ts
```diff
@@ -151,23 +151,23 @@ export class StepFunctionContextService {
...
-      const ch = hex.charAt(i);
-      binary = binary + this.hexToBinary(ch);
+    for (const num of hex) {
+      binary = binary + this.numberToBinaryString(num);
     }
 
-    const res = "0" + binary.substring(1, 64);
-    if (res === "0".repeat(64)) {
+    const res = "0" + binary.substring(1, 128);
+    if (res === "0".repeat(128)) {
```
nine5two7, line 164: Here, we need to ensure that the 0th and 64th bits of the 128-bit data are both set to 0.

nine5two7 (approved):
LGTM

src/trace/step-function-service.spec.ts
```diff
@@ -268,36 +269,44 @@ describe("StepFunctionContextService", () => {
...
+      );
+      expect(actual).toEqual("3711631873188331089");
+    });
+
+    it("test same hashing number is generated as logs-backend for execution id # state name # entered time", () => {
+      const actual = instance["deterministicSha256HashToBigIntString"](
+        "arn:aws:states:sa-east-1:425362996713:stateMachine:MyStateMachine-b276uka1j#lambda#2",
+        _64_BITS,
+      );
+      expect(actual).toEqual("5759173372325510050");
```
kimi-p, line 309: this matches logs-backend's hash result
https://github.com/DataDog/logs-backend/pull/71616/files#diff-820e0a1dd4f0f97815ac54f993c08b3c44c67dd26a312968c214f16ab73494c0R20-R21

src/trace/step-function-service.spec.ts
```diff
@@ -268,36 +269,44 @@ describe("StepFunctionContextService", () => {
...
   });
+
+  describe("test 64 bits deterministicMd5HashToBigIntString for span id", () => {
+    const instance = StepFunctionContextService.instance();
+    it("first test of #1", () => {
+      const actual = instance["deterministicSha256HashToBigIntString"](
+        "arn:aws:states:sa-east-1:425362996713:stateMachine:MyStateMachine-b276uka1j#lambda#1",
+        _64_BITS,
+      );
+      expect(actual).toEqual("3711631873188331089");
```
kimi-p, line 301: this matches logs-backend PR's hashing result
https://github.com/DataDog/logs-backend/pull/71616/files#diff-820e0a1dd4f0f97815ac54f993c08b3c44c67dd26a312968c214f16ab73494c0R15

src/trace/step-function-service.ts
```diff
@@ -123,51 +127,69 @@ export class StepFunctionContextService {
...
+    const hash = new Sha256();
+    hash.update(s);
+    const uint8Array = hash.digestSync();
+    let hex;
+    if (type === TRACE_ID) {
+      hex = uint8Array.subarray(8, 16);
+    } else {
+      // type === SPAN_ID || type === DD_P_TID
+      hex = uint8Array.subarray(0, 8);
+    }
```
kimi-p, line 178: This section matches the logic that we put on logs-backend implementation.

src/trace/step-function-service.ts
```diff
@@ -123,51 +127,68 @@ export class StepFunctionContextService {
...
+    const hash = new Sha256();
+    hash.update(s);
+    const uint8Array = hash.digestSync();
+    let intArray;
+    if (type === TRACE_ID) {
+      intArray = uint8Array.subarray(8, 16);
+    } else {
+      // type === SPAN_ID || type === DD_P_TID
+      intArray = uint8Array.subarray(0, 8);
+    }
```
kimi-p, line 178: This section matches the logic that we put on logs-backend implementation.

yarn.lock
```diff
@@ -65,400 +65,410 @@
...
-    "@smithy/util-base64" "^2.3.0"
-    "@smithy/util-body-length-browser" "^2.2.0"
-    "@smithy/util-body-length-node" "^2.3.0"
-    "@smithy/util-defaults-mode-browser" "^2.2.0"
-    "@smithy/util-defaults-mode-node" "^2.3.0"
-    "@smithy/util-endpoints" "^1.2.0"
-    "@smithy/util-middleware" "^2.2.0"
-    "@smithy/util-retry" "^2.2.0"
-    "@smithy/util-utf8" "^2.3.0"
+    "@aws-sdk/client-sso-oidc" "3.577.0"
```
nine5two7, line 84: Is it normal to have so many changes on the yarn.lock file?

kimi-p: I think it's just that these are the result of not pinned packages. When I remove the lock file and regenerated, some packages have been upgraded all at once.

src/trace/step-function-service.ts
```diff
@@ -123,51 +127,68 @@ export class StepFunctionContextService {
...
-      sampleMode,
-      source: TraceSource.Event,
+    const ddTraceContext = new _DatadogSpanContext({
+      traceId: id(traceId, 10),
+      spanId: id(parentId, 10),
+      sampling: { priority: sampleMode.toString(2) },
     });
 
+    const ptid = this.deterministicSha256HashToBigIntString(this.context["step_function.execution_id"], DD_P_TID);
+    if (ptid === "0".repeat(16)) {
```
nine5two7, line 150: Can you add some explanations for the condition check in Line 15?

kimi-p: Good catch. I think this was coming from similar logic implemented somewhere the dd-trace-js. But since `deterministicSha256HashToBigIntString` will never return all zeros, we can remove the check here.

nine5two7 (approved):
LGTM

src/trace/step-function-service.ts
```diff
@@ -123,51 +127,65 @@ export class StepFunctionContextService {
...
+    const ddTraceContext = new _DatadogSpanContext({
+      traceId: id(traceId, 10),
+      spanId: id(parentId, 10),
+      sampling: { priority: sampleMode.toString(2) },
     });
 
+    const ptid = this.deterministicSha256HashToBigIntString(this.context["step_function.execution_id"], DD_P_TID);
+    ddTraceContext._trace.tags["_dd.p.tid"] = id(ptid, 10).toString(16);
+    const spanContext = new SpanContextWrapper(ddTraceContext, TraceSource.Event);
+
```
duncanista, line 152: Why are we not using the current `SpanContextWrapper` code? Is it because we need to add the ptid?

If so, is there another way to not require `dd-trace` code here? Or at least, we need to lazy load it, since the tracer could be not available

src/trace/step-function-service.ts
```diff
@@ -123,51 +127,65 @@ export class StepFunctionContextService {
...
+    hash.update(s);
+    const uint8Array = hash.digestSync();
+    let intArray;
+    if (type === TRACE_ID) {
+      intArray = uint8Array.subarray(8, 16);
+    } else {
+      // type === SPAN_ID || type === DD_P_TID
+      intArray = uint8Array.subarray(0, 8);
+    }
+    let binaryString = "";
```
duncanista, line 176:
```suggestion
    // type === SPAN_ID || type === DD_P_TID
    let intArray = uint8Array.subarray(0, 8);
    if (type === TRACE_ID) {
      intArray = uint8Array.subarray(8, 16);
    }
    let binaryString = "";
```

src/trace/step-function-service.ts
```diff
@@ -123,51 +127,65 @@ export class StepFunctionContextService {
...
+    for (const num of intArray) {
+      binaryString = binaryString + this.numberToBinaryString(num);
     }
```
duncanista, line 180:
```suggestion
    const binaryString = intArray.reduce((acc, num) => acc + this.numberToBinaryString(num), "");

```

duncanista (commented):
Left a comment

src/trace/step-function-service.ts
kimi-p: 1. Yes.
2. Lemme try to lazy load it.

src/trace/step-function-service.ts
```diff
@@ -137,7 +137,11 @@ export class StepFunctionContextService {
       PARENT_ID,
     );
     const sampleMode = SampleMode.AUTO_KEEP;
-    const _DatadogSpanContext = require("dd-trace/packages/dd-trace/src/opentracing/span_context");
+    function getDatadogSpanContext() {
```
duncanista, line 140: Sorry, this might not cut it, we want to have a try-catch just like the wrapper method which does the same thing, I can push the change tomorrow morning if you want

kimi-p: Sure. Thanks! I guess you are worried about the customers may miss the tracer. I'm not familiar with the idea of missing tracer. Shouldn't tracer alway be included in this layer?

kimi-p:
I have verified after these change based on the comments that the spans linking still works in staging.

image

duncanista (approved):
LGTM

PR #538, opened 2024-05-06
[SVLS-4780] upstream Java sns and sqs refactor

joeyzhao2018:
### What does this PR do?
Extract trace context in SNS and SQS cases where the upstream service is instrumented by dd-trace-java
In short, 
1. For `Java => SNS => Lambda`, the env: _X_AMZN_TRACE_ID has the dd trace id but not the correct parent id. We will still use it here.
2. For `Java => SQS => Lambda` or `Java => SNS => SQS => Lambda` case, AWS propagate the trace context in the json payload attributes. 

Longer version:

javasnssqs drawio

### Testing Guidelines

- Unittests added
- Integration tests added https://github.com/DataDog/dd-trace-x-lambda-examples/blob/main/lib/wip-cases/sns-java-checks-nested-stack.ts
- Manually tested (You can see the parenting issue causing a white space between spans)
Single Message Publish
Screenshot 2024-05-09 at 1 24 45 PM

Batch Publish
Screenshot 2024-05-09 at 10 19 40 AM

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/trace/context/extractors/sns-sqs.ts
+++ b/src/trace/context/extractors/sns-sqs.ts
@@ -3,32 +3,46 @@ import { EventTraceExtractor } from "../extractor";
 import { TracerWrapper } from "../../tracer-wrapper";
 import { logDebug } from "../../../utils";
 import { SpanContextWrapper } from "../../span-context-wrapper";
+import { XrayService } from "../../xray-service";
 
 export class SNSSQSEventTraceExtractor implements EventTraceExtractor {
   constructor(private tracerWrapper: TracerWrapper) {}
 
   extract(event: SQSEvent): SpanContextWrapper | null {
-    const body = event?.Records?.[0]?.body;
-    if (body === undefined) return null;
-
     try {
-      const parsedBody = JSON.parse(body) as SNSMessage;
-      const messageAttribute = parsedBody?.MessageAttributes?._datadog;
-      if (messageAttribute.Value === undefined) return null;
+      // First try to extract trace context from message attributes
+      if (event?.Records?.[0]?.body) {
+        const parsedBody = JSON.parse(event?.Records?.[0]?.body) as SNSMessage;
+        const messageAttribute = parsedBody?.MessageAttributes?._datadog;
+        if (messageAttribute?.Value) {
+          let headers;
+          if (messageAttribute.Type === "String") {
+            headers = JSON.parse(messageAttribute.Value);
+          } else {
+            const decodedValue = Buffer.from(messageAttribute.Value, "base64").toString("ascii");
+            headers = JSON.parse(decodedValue);
+          }
 
-      let headers;
-      if (messageAttribute.Type === "String") {
-        headers = JSON.parse(messageAttribute.Value);
-      } else {
-        const decodedValue = Buffer.from(messageAttribute.Value, "base64").toString("ascii");
-        headers = JSON.parse(decodedValue);
+          const traceContext = this.tracerWrapper.extract(headers);
+          if (traceContext) {
+            logDebug("Extracted trace context from SNS-SQS event", { traceContext, event });
+            return traceContext;
+          } else {
+            logDebug("Failed to extract trace context from SNS-SQS event", { event });
+          }
+        }
       }
-
-      const traceContext = this.tracerWrapper.extract(headers);
-      if (traceContext === null) return null;
-
-      logDebug("Extracted trace context from SNS-SQS event", { traceContext, event });
-      return traceContext;
+      // Then try to extract trace context from attributes.AWSTraceHeader. (Upstream Java apps can
+      // pass down Datadog trace context in the attributes.AWSTraceHeader in SQS case)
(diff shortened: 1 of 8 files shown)


src/trace/context/extractors/sns-sqs.ts
```diff
@@ -3,32 +3,46 @@ import { EventTraceExtractor } from "../extractor";
...
+      // Then try to extract trace context from attributes.AWSTraceHeader. (Upstream Java apps can
+      // pass down Datadog trace context in the attributes.AWSTraceHeader in SQS case)
+      if (event?.Records?.[0]?.attributes?.AWSTraceHeader !== undefined) {
+        const traceContext = XrayService.extraceDDContextFromAWSTraceHeader(event.Records[0].attributes.AWSTraceHeader);
+        if (traceContext) {
+          logDebug("Extracted trace context from SNS-SQS event attributes.AWSTraceHeader", { traceContext, event });
+          return traceContext;
+        } else {
+          logDebug("No Datadog trace context found from SNS-SQS event attributes.AWSTraceHeader", { event });
+        }
```
duncanista, line 44: Is there a way to make this code more readable?

I wonder if we could create functions inside this class for specific extraction scenarios, making the `extract` look more simple. WDYT?

src/trace/context/extractors/sns.ts
```diff
@@ -3,29 +3,44 @@ import { TracerWrapper } from "../../tracer-wrapper";
...
+      // pass down Datadog trace id (parent id wrong) in the env in SNS case)
+      if (process.env[AMZN_TRACE_ID_ENV_VAR]) {
+        const traceContext = XrayService.extraceDDContextFromAWSTraceHeader(process.env[AMZN_TRACE_ID_ENV_VAR]);
+        if (traceContext) {
+          logDebug("Extracted Datadog trace context from _X_AMZN_TRACE_ID", { traceContext, event });
+          return traceContext;
+        } else {
+          logDebug("No Datadog trace context found from _X_AMZN_TRACE_ID", { event });
+        }
+      }
```
duncanista, line 43: Same as the comment above! 🤔

src/trace/context/extractors/sqs.ts
```diff
@@ -10,23 +10,28 @@ export class SQSEventTraceExtractor implements EventTraceExtractor {
...
+      // Then try to extract trace context from attributes.AWSTraceHeader. (Upstream Java apps can
+      // pass down Datadog trace context in the attributes.AWSTraceHeader in SQS case)
+      if (event?.Records?.[0]?.attributes?.AWSTraceHeader !== undefined) {
+        const traceContext = XrayService.extraceDDContextFromAWSTraceHeader(event.Records[0].attributes.AWSTraceHeader);
+        if (traceContext) {
+          logDebug("Extracted trace context from SQS event attributes AWSTraceHeader", { traceContext, event });
+          return traceContext;
+        } else {
+          logDebug("No Datadog trace context found from SQS event attributes AWSTraceHeader", { event });
+        }
```
duncanista, line 33: I think we are missing to return `null` here when we cannot extract anything, should we add it?

duncanista, line 33:
```suggestion
      // log debug "trying to extract from Xray"
      if (event?.Records?.[0]?.attributes?.AWSTraceHeader !== undefined) {
          return XrayService.extraceDDContextFromAWSTraceHeader(event.Records[0].attributes.AWSTraceHeader);
        } 
```

I wonder if this can be just modified to something like

duncanista (commented):
Left some comments.

src/trace/context/extractors/sns-sqs.ts
joeyzhao2018: I actually think it's more readable to add the logging at this level to keep it consistent as above.
The `XrayService.extraceDDContextFromAWSTraceHeader` will just try to extract dd trace context if it exists or return null. 
"Returning if found, moving on if not" is more specific at this level here.
Let me know if it makes sense?

src/trace/context/extractors/sqs.ts
joeyzhao2018: This is a "return if found, move on if not" block. A null will be returned at the end of the whole method if reached.

joeyzhao2018: Um.... I just think it's more critical to log afterwards on the result of `XrayService.extraceDDContextFromAWSTraceHeader`. It would tell us if it's found or not. And also directly returning is not intended here.

joeyzhao2018:
> Left some comments.

@duncanista 's comments made me to take a closer look the way i added debug logs. And after some thoughts I think it's best to use just one static string as the parameter and removed extra logging of the event payload or tracecontext. This is to avoid any extra memory allocation and keep the code cleaner. Based on my on-call experience, we usually get the event payload from other logs.

duncanista (approved):

Issue #545, opened 2024-05-14
sendDistributionMetricWithDate no longer including tags


pedrovanzella:
## Expected Behavior
When calling `sendDistributionMetricWithDate`, some default tags are included:

* `account_id`
* `architecture`
* `aws_account`
* `env`
* `functioname`
* `region`
* `resource`

This used to be the case up until v8.108.

## Actual Behavior

Starting with v8.109, these tags are no longer included in the metrics. Tags are still included when calling `sendDistrubtionMetric` though.

## Specifications

  - Datadog Lambda Layer version: 109
  - Node version: 18

duncanista:
Hey @pedrovanzella,

This should have been fixed in the [latest release ](https://github.com/DataDog/datadog-lambda-js/releases/tag/v9.113.0), or layer `113`.

Please feel free to re-open this issue if you have any other questions regarding this!

PR #559, opened 2024-06-28
[SVLS-4999] Add Lambda tags to metrics sent via the API

DylanLovesCoffee:
### What does this PR do?

When the Extension is present, the Lambda library will use the DD API to send distribution metrics with timestamps instead of the Extension, since the Extension does not support historical distribution metrics. Metrics sent via the API currently do not get tagged with Lambda tags. 

This adds some essential tags like those added via `DD_TAGS` and the `function_arn`. With the `function_arn` tag available, custom metrics will be enriched with other tags collected by the metadata crawler if the AWS integration is enabled:
Screenshot 2024-06-28 at 10 42 53 AM

Also added a warning log for historical distribution metrics that are sent to the API with an unsupported timestamp (>=4hrs).
Screenshot 2024-06-28 at 10 24 08 AM

### Motivation

https://github.com/DataDog/datadog-lambda-python/issues/398 and #545 

### Testing Guidelines

Built the layer and tested manually.

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -421,13 +421,13 @@ function getRuntimeTag(): string {
 
 function getDDTags(): string[] {
   const ddTags = getEnvValue("DD_TAGS", "").split(",");
-  const ddService = getEnvValue("DD_SERVICE", "")
+  const ddService = getEnvValue("DD_SERVICE", "");
   if (ddService.length > 0) {
-    ddTags.push(`service:${ddService}`)
+    ddTags.push(`service:${ddService}`);
   }
-  const ddEnv = getEnvValue("DD_ENV", "")
+  const ddEnv = getEnvValue("DD_ENV", "");
   if (ddEnv.length > 0) {
-    ddTags.push(`env:${ddEnv}`)
+    ddTags.push(`env:${ddEnv}`);
   }
   return ddTags;
 }
--- a/src/index.ts
+++ b/src/index.ts
@@ -153,7 +153,7 @@ export function datadog<TEvent, TResult>(
 
     try {
       await traceListener.onStartInvocation(event, context);
-      await metricsListener.onStartInvocation(event);
+      await metricsListener.onStartInvocation(event, context);
       if (finalConfig.enhancedMetrics) {
         incrementInvocationsMetric(metricsListener, context);
       }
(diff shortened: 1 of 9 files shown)


src/metrics/batcher.ts
```diff
@@ -23,9 +23,7 @@ export class Batcher {
    * Convert batched metrics to a list of api compatible metrics
    */
   public toAPIMetrics(): APIMetric[] {
-    return [...this.metrics.values()]
-      .map((metric) => metric.toAPIMetrics()) // No flatMap support yet in node 10
-      .reduce((prev, curr) => prev.concat(curr), []);
+    return [...this.metrics.values()].flatMap((metric) => metric.toAPIMetrics());
```
DylanLovesCoffee, line 26: small refactor

src/index.ts
```diff
@@ -419,6 +419,19 @@ function getRuntimeTag(): string {
   return `dd_lambda_layer:datadog-node${version}`;
 }
 
+function getDDTags(): string[] {
```
duncanista, line 422: Can you add testing for this function?

src/metrics/listener.ts
```diff
@@ -138,9 +142,20 @@ export class MetricsListener {
     if (this.isExtensionRunning) {
       const isMetricTimeValid = Date.parse(metricTime.toString()) > 0;
       if (isMetricTimeValid) {
+        const dateCeiling = new Date(Date.now() - 4 * 60 * 60 * 1000); // 4 hours ago
+        if (dateCeiling > metricTime) {
+          logWarning(
+            "The timestamp provided is too old to be sent to the Datadog API. Please provide a timestamp within the last 4 hours.",
+          );
```
duncanista, line 149: Would it be possible to get this text to be more specific? Something like `Metric {name} ..." in case a user is sending thousands of metrics so its easier to debug?

duncanista: Also, is there a test for us to check that a metric is indeed ignored?

DylanLovesCoffee: yessir, will add these 👍

src/index.ts
DylanLovesCoffee: Can do!

src/metrics/listener.ts
```diff
@@ -138,9 +142,18 @@ export class MetricsListener {
     if (this.isExtensionRunning) {
       const isMetricTimeValid = Date.parse(metricTime.toString()) > 0;
       if (isMetricTimeValid) {
+        const dateCeiling = new Date(Date.now() - 4 * 60 * 60 * 1000); // 4 hours ago
```
duncanista, line 145: constant?

duncanista (approved):

PR #563, opened 2024-08-01
ensure tracer initialization does not override DD_TAGS

wconti27:
### What does this PR do?

Prevents `DD_TAGS` from being overwritten.

### Motivation

Customer escalation

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/runtime/module_importer.js
+++ b/src/runtime/module_importer.js
@@ -13,8 +13,8 @@ exports.initTracer = function () {
     const path = require.resolve("dd-trace", { paths: ["/var/task/node_modules", ...module.paths] });
     // tslint:disable-next-line:no-var-requires
     // add lambda tags to DD_TAGS environment variable
-    updateDDTags("_dd.origin:lambda")
-    const tracer = require(path).init();
+    const ddtags = updateDDTags({"_dd.origin": "lambda"})
+    const tracer = require(path).init({tags: ddtags});
     logDebug("automatically initialized dd-trace");
 
     // Configure the tracer to ignore HTTP calls made from the Lambda Library to the Extension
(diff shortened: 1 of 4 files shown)


src/runtime/module_importer.js
```diff
@@ -12,11 +12,9 @@ exports.initTracer = function () {
...
     const path = require.resolve("dd-trace", { paths: ["/var/task/node_modules", ...module.paths] });
     // tslint:disable-next-line:no-var-requires
-    const tracer = require(path).init({
-        tags: {
-            "_dd.origin": "lambda",
-        },
-    });
+    // add lambda tags to DD_TAGS environment variable
+    updateDDTags("_dd.origin:lambda")
+    const tracer = require(path).init();
```
duncanista, line 17: I'm not sure this is the right approach. I'd oppose in updating the env var. At the same time, I think it leaves our team with no room to add additional config if required.

wconti27: I've changed this to simply include the env variable values in the tags config arg

src/runtime/module_importer.js
```diff
@@ -12,11 +12,9 @@ exports.initTracer = function () {
     // the version provided by the layer
     const path = require.resolve("dd-trace", { paths: ["/var/task/node_modules", ...module.paths] });
     // tslint:disable-next-line:no-var-requires
-    const tracer = require(path).init({
-        tags: {
-            "_dd.origin": "lambda",
-        },
-    });
+    // add lambda tags to DD_TAGS environment variable
+    const ddtags = updateDDTags({"_dd.origin": "lambda"})
```
duncanista, line 16: Name might differ with the current existing function used in metrics, but might be a good think to refactor in the future

duncanista (approved):

PR #585, opened 2024-10-25
Detect and warn if duplicate lambda libraries are detected

nhulston:
### What does this PR do?

Asynchronously detects if the Lambda library is installed both in node_modules and as a Lambda library. If so, it logs a warning.

### Motivation

It's common for users to install the Lambda layers as instructed by the documentation, but then then see import errors in their code, causing them to run npm install datadog-lambda-js. Now they will have the Lambda library installed in two places, which will (1) increase package size and cold start duration, and (2) break custom metrics.

We've had several customers open support tickets recently about broken metrics caused by this issue, so this PR aims to avoid this problem in the future.

### Testing Guidelines

Manual tests: logs a warning as expected when installed in both places, and doesn't log any warning if only installed in the layer or node_modules.

Added unit tests.

### Additional Notes

This works by using `fs` and checking if the paths exist. I'm assuming the paths will always be:
- Layer: `'/opt/nodejs/node_modules/datadog-lambda-js'`
- Node modules: `path.join(process.cwd(), 'node_modules/datadog-lambda-js')`

If I'm mistaken and these paths could be different, please let me know and I'll make the necessary changes.

There is probably a "better" solution if we use `process.env.NODE_PATH`, but this solution is simpler (less overhead) and works perfectly fine.

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


src/index.ts
```diff
@@ -127,6 +133,19 @@ export function datadog<TEvent, TResult>(
 
   const traceListener = new TraceListener(finalConfig);
 
+  // Check for duplicate installations of the Lambda library
+  detectDuplicateInstallations()
+    .then((duplicateFound) => {
```
nhulston, line 138: I put this under the `datadog()` wrapper, but I'm not sure if this is the best place to make this call. If there's somewhere that's better, I'm open to suggestions!

src/index.ts
```diff
@@ -131,6 +137,19 @@ export function datadog<TEvent, TResult>(
...
+      if (duplicateFound) {
```
nhulston, line 143: Also, this code might be relevant; maybe we move some of this to the trace wrapper file?
https://github.com/DataDog/datadog-lambda-js/blob/f05c6003a4f7bab82e46e741ac8d0c98d89efe5b/src/trace/tracer-wrapper.ts#L28-L44

Note that this code only loads in one version of the tracer, not to be confused with the lambda library; this PR detects duplicate installations of the lambda library.

duncanista: Wonder if there's a way for us to use similar logic like this one 🤔

PR #587, opened 2024-11-14
Add S3 Downstream Span Pointers

nhulston:
### What does this PR do?

Adds span pointers to spans for Lambdas triggered by `putObject`, `copyObject`, and `completeMultipartUpload` events.

Span pointers are similar to [Span Links](https://docs.datadoghq.com/tracing/trace_collection/span_links/), but for cases when it is impossible to pass the Trace ID and Span ID between the spans that need to be linked.

When the calculated hashes for the upstream and downstream lambdas match, the Datadog backend will automatically link the two traces together.

Screenshot 2024-11-11 at 1 29 11 PM
When clicking on the linked span, a new tab opens linking to the downstream Lambda function that was triggered by this S3 object update.

### Motivation

This feature already exists in Python, and I'm working on adding it to all other runtimes (Node, .NET, Java, Golang).

### Testing Guidelines
Easy: Checkout [this span](https://ddserverless.datadoghq.com/functions?cloud=aws&config_serverless-auto-span-links=true&entity_view=lambda_functions&fromUser=false&graphType=flamegraph&panel_end=1731688980000&panel_paused=false&panel_start=1731688080000&shouldShowLegend=true&sort=time&sp=%5B%7B%22p%22%3A%7B%22entityId%22%3A%22aws-lambda-functions%2Bnhulston-javascript-upstream-dev-hello%2Bus-east-1%2B425362996713%22%7D%2C%22i%22%3A%22lambda-panel%22%7D%2C%7B%22p%22%3A%7B%22traceID%22%3A%226833201988596926111%22%2C%22selectedSpanID%22%3A%226833201988596926111%22%7D%2C%22i%22%3A%22trace-panel%22%7D%5D&spanID=6833201988596926111&spanViewType=span-links&text_search=nhulston&traceID=6833201988596926111&start=1731685380000&end=1731688980000&paused=false), enable the feature flag, and you will see that it's pointing to the downstream Lambda.

More thorough: Run [this Lambda function](https://us-east-1.console.aws.amazon.com/lambda/home?region=us-east-1#/functions/nhulston-javascript-upstream-dev-hello?subtab=envVars&tab=testing) with the event payload
```json
{
  "shouldPutObject": false,
  "shouldCopyObject": false,
  "shouldMultipartUpload": false,
  "useV2": true
}
```
and change one of the bools to true. Also, try both with AWS SDK v2 and v3. Enable the span pointers feature flag, and check Datadog to ensure that the spans are properly linked.

### Additional Notes

- This handles the 'downstream' case, for Lambdas that make S3 requests. [This PR](https://github.com/DataDog/dd-trace-js/pull/4875) adds span pointers on the upstream case.
- S3 and Dynamo are supported in Python. I will soon open another PR to add support for Dynamo.

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


src/utils/span-pointers.ts
```diff
@@ -0,0 +1,74 @@
...
+}
+
+function processS3Event(event: any): SpanPointerAttributes[] {
+  const records = event.Records || [];
+  const spanPointerAttributesList = [];
+  const linkKind = SPAN_LINK_KIND;
+
+  for (const record of records) {
+    const eventName = record.eventName;
+    if (!["ObjectCreated:Put", "ObjectCreated:Copy", "ObjectCreated:CompleteMultipartUpload"].includes(eventName)) {
```
apiarian-datadog, line 48: why restrict us to only these particular object created types instead of `ObjectCreated:*`?

src/utils/span-pointers.ts
```diff
@@ -0,0 +1,74 @@
...
+      logDebug("Unable to calculate span pointer hash because of missing parameters.");
+      continue;
+    }
+
+    const pointerHash = generateS3PointerHash(bucketName, objectKey, eTag);
+    const spanPointerAttributes = {
+      "ptr.kind": S3_PTR_KIND,
+      "ptr.dir": SPAN_POINTER_DIRECTION.UPSTREAM,
+      "ptr.hash": pointerHash,
+      "link.kind": linkKind,
```
apiarian-datadog, line 68: a little weird that we need to tell span pointers that they're span pointers =]

apiarian-datadog (approved):
the span pointer stuff looks fine, though some aspects of the api are a little strange. we may want someone with more js experience to take a look as well.

src/utils/span-pointers.ts
nhulston: We only support creating span pointers for these three operations on the upstream case. I guess it doesn't hurt to create span pointers downstream in other cases; we could change the check to

`if (!eventName.startsWith("ObjectCreated"))`

src/utils/span-pointers.ts
nhulston: What do you mean?

src/utils/span-pointers.spec.ts
```diff
@@ -0,0 +1,159 @@
...
+  pointerKind: string;
+  pointerDirection: string;
+  pointerHash: string;
+}
+
+describe("span-pointers utils", () => {
+  const mockPointerHash = "mock-hash-123";
+
+  beforeEach(() => {
+    jest.spyOn(util, "generatePointerHash").mockReturnValue(mockPointerHash);
```
apiarian-datadog, line 24: why not let the generate pointer hash function actually run?

src/utils/span-pointers.ts
```diff
@@ -0,0 +1,85 @@
...
+): SpanPointerAttributes[] | undefined {
+  if (!eventSource) {
+    return;
+  }
+
+  switch (eventSource) {
+    case eventTypes.s3:
+      return processS3Event(event);
+    default:
+      logDebug(`Event type ${eventSource} not supported by span pointers.`);
```
apiarian-datadog, line 29: is this log line necessary?

src/utils/span-pointers.spec.ts
nhulston: I already have tests for `generatePointerHash` in the tracer; i just wanted to test only the logic behind the functions created in this PR. That way, if we ever make changes to how the `generatePointerHash` function works in the tracer, it won't break the tests here.

src/utils/span-pointers.ts
nhulston: Nope, it probably is making more noise than it would help. Good catch

src/trace/listener.ts
```diff
@@ -70,6 +71,12 @@ export interface TraceConfig {
   coldStartTraceSkipLib: string;
 }
 
+interface SpanPointerAttributes {
+  pointerKind: string;
+  pointerDirection: string;
+  pointerHash: string;
+}
```
purple4reina, line 78: This interface is defined in two different files. Is there a way we can import it from one to the other?

src/utils/span-pointers.ts
```diff
@@ -0,0 +1,82 @@
...
+function processS3Event(event: any): SpanPointerAttributes[] {
+  const records = event.Records || [];
+  const spanPointerAttributesList: SpanPointerAttributes[] = [];
+
+  // Get dependencies from tracer only when needed
+  let constants;
+  let util;
+  try {
+    constants = require("dd-trace/packages/dd-trace/src/constants");
+    util = require("dd-trace/packages/dd-trace/src/util");
```
purple4reina, line 40: @duncanista, can you help answer a question for me?

I recall when you added support for w3c trace propagation to datadog-lambda-js, you said that instead of directly importing the tracer from dd-trace, we create an interface to wrap it. 

If that's correct, then is there any issue with us requiring other files from ddtrace here?

duncanista: There might be an issue when customer is not using the tracer (some do it), it would essentially crash, ideally, we'd have a better way to get this type of data

src/trace/listener.ts
```diff
@@ -131,8 +139,14 @@ export class TraceListener {
...
     this.lambdaSpanParentContext = this.inferredSpan?.span || parentSpanContext;
     this.context = context;
-    this.triggerTags = extractTriggerTags(event, context);
+    const eventSource = parseEventSource(event);
+    this.triggerTags = extractTriggerTags(event, context, eventSource);
     this.stepFunctionContext = StepFunctionContextService.instance().context;
+
+    const result = getSpanPointerAttributes(eventSource, event);
+    if (result) {
+      this.spanPointerAttributesList.push(...result);
```
purple4reina, line 148: I'm not completely sure how our wrapping works, but I just want to confirm. It looks to me like `this.spanPointerAttributesList` is an instance attribute. Yet because we only create a single trace listener, then this list is never reset, and will just grow in size.

I'm also wondering, separately, if instead of calling `.push` on the list if we can just straight up set `this.spanPointerAttributesList = getSpanPointerAttributes(eventSource, event);`?

nhulston: You're right, thank you for catching this. Yes, we should be able to just set `this.spanPointerAttributesList` to the result.

src/utils/span-pointers.ts
nhulston: Could you take another look at the updated code from my latest commit? It should be safe and avoid any crashing:
```typescript
  let S3_PTR_KIND;
  let SPAN_POINTER_DIRECTION;
  let generatePointerHash;
  try {
    const constants = require("dd-trace/packages/dd-trace/src/constants");
    const util = require("dd-trace/packages/dd-trace/src/util");

    ({ S3_PTR_KIND, SPAN_POINTER_DIRECTION } = constants);
    ({ generatePointerHash } = util);
  } catch (err) {
    if (err instanceof Error) {
      logDebug("Failed to load dd-trace span pointer dependencies", err);
    }
    return spanPointerAttributesList;
  }
```

src/trace/listener.ts
```diff
@@ -131,8 +133,14 @@ export class TraceListener {
...
     this.lambdaSpanParentContext = this.inferredSpan?.span || parentSpanContext;
     this.context = context;
-    this.triggerTags = extractTriggerTags(event, context);
+    const eventSource = parseEventSource(event);
+    this.triggerTags = extractTriggerTags(event, context, eventSource);
     this.stepFunctionContext = StepFunctionContextService.instance().context;
+
+    const result = getSpanPointerAttributes(eventSource, event);
+    if (result) {
+      this.spanPointerAttributesList = result;
```
purple4reina, line 142: If this same listener instance is used each time, then I'm worried about the case where the function receives an S3 event (which gets a span pointer) then on the next invocation gets something else. Because we're doing `if (result)` the span pointers will still be present from the previous invocation. Please correct me if I'm wrong though! I think it's probably just better all around to not do the falsy check on the `result`. That's what the step function call a line above is doing.

src/utils/span-pointers.ts
purple4reina: @duncanista, Can you give more details (or point me to a doc somewhere) about how "customer is not using the tracer" is implemented? Do you mean they just aren't including `ddtrace` in their package.json file? I'd like to be able to create a small sample app that I can test this PR with that "is not using the tracer".

duncanista: Yeah correct, just the lambda library, not the tracer. Let me review the whole PR

src/utils/span-pointers.ts
```diff
@@ -0,0 +1,83 @@
+import { eventTypes } from "../trace/trigger";
+import { logDebug } from "./log";
+
+export interface SpanPointerAttributes {
+  pointerKind: string;
+  pointerDirection: string;
+  pointerHash: string;
```
duncanista, line 7: Why the need of prefixing it with `pointer` if it's already `SpanPointerAttributes?`

duncanista (approved):
This should be fine – I'm approving to unblock, but this is not ideal on how we are importing it. I guess its fail-safe, yet this is not making it friendly to eventually make this package ESBuild compatible

duncanista:
Please add integration tests for this, due to how you are requiring this packages from the external package

src/trace/listener.ts
nhulston: Yes you're correct 🤦. Thank you for catching this

src/utils/span-pointers.ts
nhulston: Good point

purple4reina (approved):
looks great!

PR #588, opened 2024-11-15
Step Functions Trace Merging Support

avedmala:
## What does this PR do?

Allows us to do multi-level trace merging between Step Functions and Node Lambda functions

### multi-level trace merging

Screenshot 2024-11-15 at 3 00 58 PM

Added logic to extract trace context from `_datadog` for Step Functions cases where...
1. Root of the overall execution is a Lambda. This means we have the `x-datadog-trace-id` and `x-datadog-tags` but we need to compute the parentID using parent SFN context object
2. Root of the overall execution is another Step Function. This means we need to compute both traceID and parentID from `RootExecutionId` and the parent SFN context object
3. Legacy case where there's no `_datadog` and we instead figure out traceID and parentID from the parent SFN context object

### live testing
1. `Lambda1 --> SFN --> Lambda2` [trace](https://dd.datad0g.com/apm/trace/673b667c000000001440b57672af3118?graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=5799951521693525512&timeHint=1731946108273.9214)
  a. Lambda1 is actually in Python, cool how we can mix and match runtimes
2. `SFN --> Lambda` [trace](https://dd.datad0g.com/apm/trace/7a46fcde4fa42b1b22dc17a341489db3?graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=3457008248562252968&timeHint=1731945399475)

### trigger tag improvements
Also fixed the trigger tags parsing for our use case!
Screenshot 2024-11-18 at 10 41 44 AM

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/trace/step-function-service.ts
+++ b/src/trace/step-function-service.ts
@@ -3,22 +3,56 @@ import { SampleMode, TraceSource } from "./trace-context-service";
 import { SpanContextWrapper } from "./span-context-wrapper";
 import { Sha256 } from "@aws-crypto/sha256-js";
 
-export interface StepFunctionContext {
-  "step_function.execution_name": string;
-  "step_function.execution_id": string;
-  "step_function.execution_input": object;
-  "step_function.execution_role_arn": string;
-  "step_function.execution_start_time": string;
-  "step_function.state_machine_name": string;
-  "step_function.state_machine_arn": string;
-  "step_function.state_entered_time": string;
-  "step_function.state_name": string;
-  "step_function.state_retry_count": number;
+interface StepFunctionRootContext {
+  execution_id: string;
+  state_entered_time: string;
+  state_name: string;
+  root_execution_id: string;
+  serverless_version: string;
 }
+
+interface LambdaRootContext {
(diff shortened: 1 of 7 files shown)


src/trace/step-function-service.ts
```diff
@@ -3,23 +3,57 @@ import { SampleMode, TraceSource } from "./trace-context-service";
 import { SpanContextWrapper } from "./span-context-wrapper";
 import { Sha256 } from "@aws-crypto/sha256-js";
 
-export interface StepFunctionContext {
```
joeyzhao2018, line 6: since the context is also used as `tags` here [code line](https://github.com/DataDog/datadog-lambda-js/blob/8087c872b1806b27c2d6cface2f1f6b4b702815c/src/trace/listener.ts#L289) can we clarify why the change of the tags in the PR description?

joeyzhao2018 (approved):
LGTM

PR #590, opened 2024-11-20
chore: upgrade `dd-trace` to `v4.50.0`

nhulston:
### What does this PR do?

Upgrades `dd-trace` to `v4.50.0`

### Motivation

Load in new span pointer features: https://github.com/DataDog/dd-trace-js/pull/4875

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


avedmala (approved):

scripts/check_layer_size.sh
```diff
@@ -9,7 +9,7 @@
 
 # 7 mb size limit
 MAX_LAYER_COMPRESSED_SIZE_KB=$(expr 7 \* 1024)
-MAX_LAYER_UNCOMPRESSED_SIZE_KB=$(expr 18 \* 1024)
+MAX_LAYER_UNCOMPRESSED_SIZE_KB=$(expr 19 \* 1024)
```
purple4reina, line 12: Just as an aside. What's the point in checking the layer size if we're just gonna keep upping the limit?

integration_tests/snapshots/logs/process-input-traced_node16.log
```diff
@@ -38,7 +38,6 @@ START
           "_dd.p.tid": "XXXX",
           "_dd.p.dm": "-0",
           "service": "remappedApiGatewayServiceName",
-          "version": "1.0.0",
```
purple4reina, line 41: Did we figure out why this was removed?

scripts/check_layer_size.sh
nhulston: Great question, I'm not sure. I guess to make sure we don't accidentally increase the size significantly without realizing?

integration_tests/snapshots/logs/process-input-traced_node16.log
```diff
@@ -48,7 +47,6 @@ START
           "_inferred_span.tag_source": "self",
           "_inferred_span.synchronicity": "sync",
           "http.method": "GET",
-          "http.route":"/{proxy+}",
```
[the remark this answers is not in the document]
nhulston: At the time of creating https://github.com/DataDog/datadog-lambda-js/pull/584
I still had the issue of not being able to update the snapshots. So I added the original lines manually, and I must have put them in the wrong place even though the tests still passed.

It seems like they _should_ be in the invocation span, not the inferred span.

integration_tests/snapshots/logs/process-input-traced_node16.log
nhulston: Not sure. Looks like it's just removed on the inferred span, but not the aws.lambda spans.

purple4reina: Okay, that sounds fine then. I think it'll be important to make sure they remain on the aws.lambda span even if they're removed from the inferred span.

integration_tests/snapshots/logs/process-input-traced_node16.log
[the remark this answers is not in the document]
purple4reina: I disagree. I think it should be on the inferred span since for an api.gateway span, that's where the idea of the routing is happening. Can we run these again and confirm where these tags are actually being added? This is the only trace out of all the test invocates in which it's been moved.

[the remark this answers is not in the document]
nhulston: The code that sets the http.route tags is here: https://github.com/DataDog/datadog-lambda-js/blob/56965ac722d419af55608f445e06f5aaadbfd858/src/trace/trigger.ts#L299-L318

If you backtrack the call stack, it starts at `onStartInvocation` and stores the result in `this.triggerTags`, so at least according to the code, we are setting the tags on the Lambda span -- not the API GW inferred span.

More info: https://datadoghq.atlassian.net/browse/SVLS-5779

purple4reina (approved):

PR #592, opened 2024-11-20
fix: Duplicate Payload Object in Input from SFN

avedmala:
### What does this PR do?

Adds a guard clause around the code where we unpack the `Payload` object if it exists in the Step Functions event. There's an edge case where it isn't a Legacy Lambda object but customer's data may have a `Payload` in there

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/trace/step-function-service.ts
+++ b/src/trace/step-function-service.ts
@@ -76,7 +76,10 @@ export class StepFunctionContextService {
     if (typeof event !== "object") return;
 
     // Extract Payload if available (Legacy lambda parsing)
-    if (typeof event.Payload === "object") {
+    if (
+      typeof event.Payload === "object" &&
+      (typeof event.Payload._datadog === "object" || this.isValidContextObject(event.Payload))
+    ) {
       event = event.Payload;
     }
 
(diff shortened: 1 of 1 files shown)


src/trace/step-function-service.ts
```diff
@@ -76,7 +76,10 @@ export class StepFunctionContextService {
     if (typeof event !== "object") return;
 
     // Extract Payload if available (Legacy lambda parsing)
-    if (typeof event.Payload === "object") {
+    if (
+      typeof event.Payload === "object" &&
+      (typeof event.Payload._datadog === "object" || this.isValidContextObject(event.Payload))
```
nhulston, line 81: Nit: We could make this more by adding optional chaining, i.e.

`typeof event?.Payload === "object"`.

Unless we already know for sure that event is defined, then we don't have to do this

src/trace/step-function-service.ts
```diff
@@ -199,27 +202,28 @@ export class StepFunctionContextService {
...
-      typeof state.Name === "string"
-    ) {
+    if (this.isValidContextObject(event)) {
       return {
-        execution_id: execution.Id,
-        state_entered_time: state.EnteredTime,
-        state_name: state.Name,
+        execution_id: event.Execution.Id,
+        state_entered_time: event.State.EnteredTime,
+        state_name: event.State.Name,
```
nhulston, line 209: same micro nit here, might be safer to do `event?.Execution?.Id` and so on

src/trace/step-function-service.ts
```diff
@@ -199,27 +202,28 @@ export class StepFunctionContextService {
...
   }
 
+  private isValidContextObject(context: any): boolean {
+    return (
+      typeof context.Execution === "object" &&
+      typeof context.Execution.Id === "string" &&
+      typeof context.State === "object" &&
+      typeof context.State.EnteredTime === "string" &&
+      typeof context.State.Name === "string"
+    );
```
nhulston, line 224: Same micro nit. Ignore me if you know 100% these fields always exist

nhulston (approved):

PR #622, opened 2025-02-18
Update Step Functions Parent ID Generation

avedmala:
Updates the parent ID generation using upstream Step Functions context object to be able to handle retries

Old behavior would result in the same parent ID for all retries, this new behavior will use the `retry_count` when it's non-zero. This way we don't break backwards compatibility with old layers that don't use this value in the common case of `redrive_count:0`

Related: https://github.com/DataDog/datadog-lambda-python/pull/559

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


nine5two7 (approved):
LGTM

PR #634, opened 2025-03-10
Use FIPs endpoint for SecretManager when in Govcloud region

nhulston:
### What does this PR do?

When creating the `SecretsManager` client, set `useFipsEndpoint` to true if in a govcloud region.

Will get to KMS key storage in a following PR.

### Motivation

Making our products FIPs compliant.

### Testing Guidelines

Manually. Secrets manager works in both Govcloud and commercial regions.

Added a simple unit test

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/metrics/listener.ts
+++ b/src/metrics/listener.ts
@@ -225,15 +225,10 @@ export class MetricsListener {
       try {
         const { SecretsManager } = await import("@aws-sdk/client-secrets-manager");
         const region = process.env.AWS_REGION;
-        let secretsParams: SecretsManagerClientConfig = {};
-        if (region) {
-          const isGovRegion = region.startsWith("us-gov-");
-          secretsParams = {
-            useFipsEndpoint: isGovRegion,
-            region,
-          };
-        }
-        const secretsManager = new SecretsManager(secretsParams);
+        const isGovRegion = region !== undefined && region.startsWith("us-gov-");
+        const secretsManager = new SecretsManager({
+          useFipsEndpoint: isGovRegion,
+        });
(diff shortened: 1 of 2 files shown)


src/metrics/listener.ts
```diff
@@ -223,7 +224,16 @@ export class MetricsListener {
...
+        const region = process.env.AWS_REGION;
+        let secretsParams: SecretsManagerClientConfig = {};
+        if (region) {
+          const isGovRegion = region.startsWith("us-gov-");
+          secretsParams = {
+            useFipsEndpoint: isGovRegion,
+            region,
+          };
+        }
+        const secretsManager = new SecretsManager(secretsParams);
```
duncanista, line 236:
```suggestion
        const region = process.env.AWS_REGION;
        const secretsParams: SecretsManagerClientConfig = {
            region,
            useFipsEndpoint: region.startsWith("us-gov-");
        };
        const secretsManager = new SecretsManager(secretsParams);
```

Do you really need to pass the region here? Or doesn't it work it you don't pass it? Normally region is used by the AWS SDK by default from the env AFAIK, just trying to figure out if you really need it? Or is it bc since you're already parsing it?

`AWS_REGION` should always exists

nhulston: You're right, we don't need to pass region. 

tested manually on commercial+gov and works fine

duncanista (approved):

hghotra (commented):
👍🏼

PR #635, opened 2025-03-11
Use FIPs endpoint for KMS when in Govcloud region

nhulston:
### What does this PR do?

When creating the `kmsClient`, set `useFipsEndpoint` to true if in a govcloud region.

### Motivation

Making our products FIPs compliant.

### Testing Guidelines

Manually. Works in both Govcloud and commercial regions.

### Additional Notes

Related PR: https://github.com/DataDog/datadog-lambda-js/pull/634

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/metrics/kms-service.ts
+++ b/src/metrics/kms-service.ts
@@ -12,6 +12,16 @@ export class KMSService {
     const buffer = Buffer.from(ciphertext, "base64");
     let kms;
 
+    const region = process.env.AWS_REGION;
+    const isGovRegion = region !== undefined && region.startsWith("us-gov-");
+    let kmsClientParams = {}
+    if (isGovRegion) {
+      // Endpoints: https://docs.aws.amazon.com/general/latest/gr/kms.html
+      kmsClientParams = {
+        endpoint: `https://kms-fips.${region}.amazonaws.com`,
+      }
+    }
+
     // Explicitly try/catch this require to appease esbuild and ts compiler
     // otherwise users would need to mark this as `external`
     // see https://github.com/DataDog/datadog-lambda-js/pull/409
@@ -20,11 +30,12 @@ export class KMSService {
     } catch (err) {
       if ((err as any).code === "MODULE_NOT_FOUND") {
         // Node 18
-        return this.decryptV3(buffer);
+        return this.decryptV3(buffer, kmsClientParams);
       }
     }
     try {
-      const kmsClient = new kms();
+      // Configure KMS client to use FIPS endpoint
+      const kmsClient = new kms(kmsClientParams);
 
       // When the API key is encrypted using the AWS console, the function name is added as an encryption context.
       // When the API key is encrypted using the AWS CLI, no encryption context is added.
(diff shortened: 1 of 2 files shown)


src/metrics/kms-service.ts
```diff
@@ -50,7 +61,7 @@ export class KMSService {
   }
 
   // Node 18 or AWS SDK V3
-  public async decryptV3(buffer: Buffer): Promise<string> {
+  public async decryptV3(buffer: Buffer, kmsClientParams: any): Promise<string> {
```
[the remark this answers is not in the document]
nhulston: we're importing the `KmsClient` with `require` so it doesn't have a type. So specifying a type other than `any` doesn't matter

src/metrics/kms-service.ts
```diff
@@ -12,6 +12,16 @@ export class KMSService {
...
     let kms;
 
+    const region = process.env.AWS_REGION;
+    const isGovRegion = region !== undefined && region.startsWith("us-gov-");
+    let kmsClientParams = {};
+    if (isGovRegion) {
+      // Endpoints: https://docs.aws.amazon.com/general/latest/gr/kms.html
+      kmsClientParams = {
+        endpoint: `https://kms-fips.${region}.amazonaws.com`,
+      };
```
duncanista, line 22:
```suggestion
      kmsClientParams['endpoint'] = `https://kms-fips.${region}.amazonaws.com`;
```

You already created the object.

nhulston: yeah but typescript. Then I'd have to do:
```typescript
    const kmsClientParams: { endpoint?: string; } = {};
    if (isGovRegion) {
      // Endpoints: https://docs.aws.amazon.com/general/latest/gr/kms.html
      kmsClientParams.endpoint = `https://kms-fips.${region}.amazonaws.com`;
    }
```

is that better performance-wise?

duncanista: Can't you just type it with the actual expected parameters type from KMS?

nhulston: No because we're importing `kmsClient` via `require` so we don't have the types.

I also see this at the top of the file
```
// in the Lambda environment anyway), we use require to import the SDK.
```

so we can't import anything to have those types

nhulston: I do think this is best because it 'works' and after TS is transpiled down to JS, we no longer create a new object
```typescript
    const kmsClientParams: { endpoint?: string; } = {};
    if (isGovRegion) {
      // Endpoints: https://docs.aws.amazon.com/general/latest/gr/kms.html
      kmsClientParams.endpoint = `https://kms-fips.${region}.amazonaws.com`;
    }
```

src/metrics/kms-service.ts
```diff
@@ -12,6 +14,19 @@ export class KMSService {
     const buffer = Buffer.from(ciphertext, "base64");
     let kms;
 
+    const region = process.env.AWS_REGION;
+    const isGovRegion = region !== undefined && region.startsWith("us-gov-");
+    if (isGovRegion) {
+      logDebug("Govcloud region detected. Using FIPs endpoints for secrets management.")
+    }
```
nhulston, line 21: this should probably be an `info` level debug, but there's only `debug`, `warn`, or `error` log levels?

duncanista (approved):

PR #639, opened 2025-03-13
Use correct SecretsManager region

nhulston:
### What does this PR do?

Users could be storing their secret in SecretsManager in a different region than the currently running Lambda. This parses the arn for the correct region.

Before this change, users would get error: `Secrets Manager can't find the specified secret.`

### Motivation

Backports https://github.com/DataDog/datadog-lambda-extension/pull/604

### Testing Guidelines

Unit tests

- Tested manually with secret in same region, and it works
- Tested manually with secret in different region, and it works

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/metrics/listener.ts
+++ b/src/metrics/listener.ts
@@ -224,7 +224,7 @@ export class MetricsListener {
     if (config.apiKeySecretARN !== "") {
       try {
         const { SecretsManager } = await import("@aws-sdk/client-secrets-manager");
-        const secretRegion = config.apiKeySecretARN.split(':')[3];
+        const secretRegion = config.apiKeySecretARN.split(":")[3];
         const lambdaRegion = process.env.AWS_REGION;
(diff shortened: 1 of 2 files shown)


src/metrics/listener.ts
```diff
@@ -224,10 +224,12 @@ export class MetricsListener {
     if (config.apiKeySecretARN !== "") {
       try {
         const { SecretsManager } = await import("@aws-sdk/client-secrets-manager");
-        const region = process.env.AWS_REGION;
-        const isGovRegion = region !== undefined && region.startsWith("us-gov-");
+        const secretRegion = config.apiKeySecretARN.split(":")[3];
```
nhulston, line 227: we don't need to do error handling here because it's already wrapped in try/catch

duncanista (approved):
Perfect, thanks for handling this!

Issue #641, opened 2025-03-18
statusCode ignored when returned as string


thingie:
Hi,
we run into an issue where we return a `{ statusCode: String(200) }` from our lambda handler wrapped in the datadog layer, it just gets ignored and doesn't get attached to the span as an attribute as we'd expect. Now we are not quite sure what is the expected behavior, but given we have no other problems with API gateway or any other AWS service, we went to inspect the datadog code and found the following line:

https://github.com/DataDog/datadog-lambda-js/blob/2f9c1e82073fef5b345babde365adfa25c555695/src/trace/trigger.ts#L394

Which doesn't make the intent on the dd side clear either, if the `resultStatusCode` is string already and not number, the function would just not do any explicit return. This feels to be at the odd with the code comment above the initial if statement and also the code construct is bit misleading at first sight.

Could this be clarified or fixed please? Thanks.

### Expected Behavior

`statusCode` typed as string gets attached to the lambda handler execution span as the `http.status_code` attribute.

### Actual Behavior

It gets silently dropped.

nhulston:
Thanks for opening this! This will be fixed in https://github.com/DataDog/datadog-lambda-js/pull/642 and out in the next release

PR #642, opened 2025-03-19
Extract status code when returned as string

nhulston:
### What does this PR do?

Handles case where the status code is a string instead of a number.
We extract this status code in `extractHTTPStatusCodeTag()` 

### Motivation

https://github.com/DataDog/datadog-lambda-js/issues/641

### Testing Guidelines

Manual+unit

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/trace/trigger.ts
+++ b/src/trace/trigger.ts
@@ -395,6 +395,8 @@ export function extractHTTPStatusCodeTag(
     // Type check the statusCode if available
     if (typeof resultStatusCode === "number") {
       return resultStatusCode.toString();
+    } else if (typeof resultStatusCode === "string") {
+      return resultStatusCode;
     }
   } else {
     return "200";
--- a/src/trace/trigger.ts
+++ b/src/trace/trigger.ts
@@ -395,9 +395,8 @@ export function extractHTTPStatusCodeTag(
     // Type check the statusCode if available
     if (typeof resultStatusCode === "number") {
       return resultStatusCode.toString();
-    } else if (typeof resultStatusCode === "string") {
-      return resultStatusCode;
     }
+    return resultStatusCode.toString();
   } else {
     return "200";
   }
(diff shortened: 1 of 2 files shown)


src/trace/trigger.ts
```diff
@@ -395,6 +395,8 @@ export function extractHTTPStatusCodeTag(
     // Type check the statusCode if available
     if (typeof resultStatusCode === "number") {
       return resultStatusCode.toString();
+    } else if (typeof resultStatusCode === "string") {
```
astuyve, line 398: probably don't need this `else` clause here at all, just return the resultStatusCode on 399.

nhulston: Sounds good. I am adding `toString()` in that case for type safety, since `resultStatusCode` is of type `any`

src/trace/trigger.ts
```diff
@@ -396,6 +396,7 @@ export function extractHTTPStatusCodeTag(
     if (typeof resultStatusCode === "number") {
       return resultStatusCode.toString();
     }
+    return resultStatusCode.toString();
```
astuyve, line 399: in this case it's already a string, so you can just return the status code itself.

src/trace/trigger.ts
```diff
@@ -396,6 +396,7 @@ export function extractHTTPStatusCodeTag(
     if (typeof resultStatusCode === "number") {
       return resultStatusCode.toString();
     }
```
joeyzhao2018, line 398:
```suggestion
```

src/trace/trigger.ts
joeyzhao2018: let's just always return `resultStatusCode.toString();` if it exists

src/trace/trigger.ts
joeyzhao2018: remove it as it doesn't do anything special for numbers

nhulston: oh yeah that's the simplest solution

joeyzhao2018 (approved):
LGTM

Issue #638, opened 2025-03-13
Silently dropped metrics when using sendDistributionMetric


AaronMoat:
## Description

I have a Lambda function, without log forwarding or the lambda extension set up, that issues a high count of metrics with `sendDistributionMetric`. 

I have noticed that, at times, some subset of a Lambda invocation's metrics are silently dropped. I tried running with `{ shouldRetryMetrics: true, debugLogging: true }`, with no useful output from the debugger to distinguish successful runs and unsuccessful runs.

Upon looking into this repo's code, I noticed that https://github.com/DataDog/datadog-lambda-js/commit/b20c40e95703a7029567d79d91996e97e3ca2cad introduced a change where we continually recreate `this.currentProcessor`, meaning that when we await it during flushing at the end of the lambda invocation, it's only awaiting one of the calls to `sendDistributionMetric`, not all of them. This could be causing my problem.

Does this sound correct?

## Specifications

  - Datadog Lambda Layer version: N/A
  - Node version: nodejs20.x

astuyve:
Hi @AaronMoat thanks for the note!

Would you mind opening a support ticket with us so we could take a closer look? You can use the in-app chat or even just email `support@datadoghq.com`. Please include this issue in the report as to help expedite the process.

Thanks!

alexgallotta:
Thanks for reporting this @AaronMoat 
We were able to replicate the issue and your intuition about the bug is correct!

luisabtrace:
Is there any suggested workaround? We're being affected by this as well, it's very frustrating.

DylanLovesCoffee:
@luisabtrace I believe we may have fixed this issue in the latest Extension and are in the process of validating this. Could you upgrade to the latest Extension version and also open a [support ticket](https://docs.datadoghq.com/help/)?

luisabtrace:
We've updated `datadog-lambda-js` to the latest version and that seems to have fixed the issue, thanks.

PR #643, opened 2025-03-25
Sles 2151 fix race condition flush

alexgallotta:
- add warning in case a failed send will drop metrics received during the try
- do not create one promise per metric. This would let the process to wait for only one promise and avoid missing metrics https://github.com/DataDog/datadog-lambda-js/issues/638

### DIFF
--- a/src/metrics/listener.ts
+++ b/src/metrics/listener.ts
@@ -146,29 +146,30 @@ export class MetricsListener {
     forceAsync: boolean,
     ...tags: string[]
   ) {
-    if (this.isExtensionRunning) {
-      const isMetricTimeValid = Date.parse(metricTime.toString()) > 0;
-      if (isMetricTimeValid) {
-        const dateCeiling = new Date(Date.now() - HISTORICAL_METRICS_THRESHOLD_HOURS); // 4 hours ago
-        if (dateCeiling > metricTime) {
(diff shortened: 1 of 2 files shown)


src/metrics/processor.ts
```diff
@@ -77,6 +77,12 @@ export class Processor {
       } catch {
         // Failed to send metrics, keep the old batch alive if retrying is enabled
         if (this.shouldRetryOnFail) {
+          const metricsReceivedWhileSending = this.batcher.toAPIMetrics();
+          if (metricsReceivedWhileSending.length > 0) {
+            logWarning(
+              `Failed to send metrics to Datadog, retrying at next interval. ${metricsReceivedWhileSending.length} netrics received will be lost`,
```
duncanista, line 83:
```suggestion
              `Failed to send metrics to Datadog, retrying at next interval. ${metricsReceivedWhileSending.length} metrics received will be lost`,
```

src/metrics/listener.ts
```diff
@@ -156,7 +156,9 @@ export class MetricsListener {
         }
         // Only create the processor to submit metrics to the API when a user provides a valid timestamp as
         // Dogstatsd does not support timestamps for distributions.
-        this.currentProcessor = this.createProcessor(this.config, this.apiKey);
+        if (this.currentProcessor === undefined) {
+          this.currentProcessor = this.createProcessor(this.config, this.apiKey);
+        }
```
DylanLovesCoffee, line 161: great catch 😅

Issue #603, opened 2024-12-20
Extract trace context when lambda is called through ALB


kimmoahokas:
## Expected Behavior

Lambda function called by AWS Application Load Balancer connects traces properly to upstream service

## Actual Behavior

Traces are not connected

## Steps to Reproduce the Problem

  1. Configure traced service to call AWS ALB
  1. Configure ALB to call lambda
  1. Instrument the lambda with datadog-lambda-js

## Specifications

  - Datadog Lambda Layer version: arn:aws:lambda:eu-west-1:464622532012:layer:Datadog-Node20-x:115
  - Node version: Node 20
  - Datadog Lambda Extension version: arn:aws:lambda:eu-west-1:464622532012:layer:Datadog-Extension-ARM:65

The event format is documented at https://docs.aws.amazon.com/lambda/latest/dg/services-alb.html

purple4reina:
Hey @kimmoahokas, thanks for reaching out. I'm curious though, from my understanding of ALB (you can see an example event [here](https://docs.aws.amazon.com/lambda/latest/dg/services-alb.html)) our extractors should be able to handle them out of the box. We even have a test for this [here](https://github.com/DataDog/datadog-lambda-js/blob/main/event_samples/application-load-balancer.json). Basically, whenever there's a `header` key in the event payload, we'll look through it when extracting trace context.

Instead, I'm wondering if your issue is upstream. We should first check to make sure the caller is correctly injecting trace context. You can check this by inspecting the inbound lambda payload by adding `DD_CAPTURE_LAMBDA_PAYLOAD=true` and then looking at the `aws.lambda` span on any trace. Would you mind doing that and sharing an example inbound event payload with us? (of course redact anything sensitive)

kimmoahokas:
Hello @purple4reina oh, thanks for the suggestion. I'm sure our upstream component adds trace headers (It's scala play play_ws client). But I will investigate this more after the holidays.

Mostly as a reminder for myself, this particular ALB+Lambda combination uses [ALB Multi-Value headers](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html#multi-value-headers) which apparently changes the incomign event structure. So that might be the culprit here. Anyway, I'll get back to this on January since this is not a critical problem for us at the moment.

purple4reina:
[...]

PR #647, opened 2025-04-29
Extract trace context from ALB multiValueHeaders

nhulston:
### What does this PR do?

Support trace extraction when we receive `multiValueHeaders` instead of `headers`. This can happen if multi value headers are turned on in ALB.

The event payload structure for multiValueHeaders is defined here: https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html

### Motivation

https://github.com/DataDog/datadog-lambda-js/issues/603
https://datadoghq.atlassian.net/browse/SVLS-6725

### Testing Guidelines

Manually, unit tests

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/trace/context/extractors/http.spec.ts
+++ b/src/trace/context/extractors/http.spec.ts
@@ -97,6 +97,36 @@ describe("HTTPEventTraceExtractor", () => {
       expect(traceContext?.source).toBe("event");
     });
 
+    it("extracts trace context from payload with multiValueHeaders", () => {
+      mockSpanContext = {
+        toTraceId: () => "123",
+        toSpanId: () => "456",
+        _sampling: { priority: "1" },
+      };
+      const tracerWrapper = new TracerWrapper();
+      const payload = {
+        multiValueHeaders: {
+          "X-Datadog-Trace-Id":      ["123", "789"],
+          "X-Datadog-Parent-Id":     ["456"],
+          "X-Datadog-Sampling-Priority": ["1"],
+        },
+      };
+      const extractor = new HTTPEventTraceExtractor(tracerWrapper);
+      const traceContext = extractor.extract(payload);
+
+      expect(traceContext).not.toBeNull();
+      expect(spyTracerWrapper).toHaveBeenCalledWith({
+        "x-datadog-trace-id":      "123",
+        "x-datadog-parent-id":     "456",
+        "x-datadog-sampling-priority": "1",
+      });
+
+      expect(traceContext?.toTraceId()).toBe("123");
+      expect(traceContext?.toSpanId()).toBe("456");
+      expect(traceContext?.sampleMode()).toBe("1");
+    });
+
+
     it("extracts trace context from payload with authorizer", () => {
       mockSpanContext = {
         toTraceId: () => "2389589954026090296",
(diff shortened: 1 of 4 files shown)


purple4reina (approved):

src/trace/context/extractors/http.spec.ts
```diff
@@ -97,6 +97,35 @@ describe("HTTPEventTraceExtractor", () => {
       expect(traceContext?.source).toBe("event");
     });
 
+    it("extracts trace context from payload with multiValueHeaders", () => {
+      mockSpanContext = {
```
duncanista, line 101: Could you add an event from either ALB or API GW with real values to ensure this works? That way we can re-use it in Bottlecap too

src/trace/context/extractors/http.ts
```diff
@@ -41,11 +41,17 @@ export class HTTPEventTraceExtractor implements EventTraceExtractor {
...
+    const headers = event.headers ?? event.multiValueHeaders;
     const lowerCaseHeaders: { [key: string]: string } = {};
 
-    for (const key of Object.keys(headers)) {
-      lowerCaseHeaders[key.toLowerCase()] = headers[key];
+    for (const [key, val] of Object.entries(headers)) {
+      if (Array.isArray(val)) {
+        // MultiValueHeaders: take the first value
+        lowerCaseHeaders[key.toLowerCase()] = val[0] ?? "";
+      } else if (typeof val === "string") {
```
BridgeAR, line 51: Could this ever be false? If yes, should it really be ignored?

nhulston: Looks like HTTP headers are always strings, and multiValueHeaders are always arrays of strings
https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html

BridgeAR: I would likely just remove the if check in that case while you can keep it to be on the very safe side.

nhulston: True, although typescript gets angry if i remove it. i think it's fine to be on the safe side

src/trace/context/extractors/http.spec.ts
nhulston: Just commenting for visibility -- as discussed, this change only applies to ALB. API GW can have multiValueHeaders, but trace context is always injected under the regular `headers` so no changes needed for API GW

nhulston: But I added a real event from ALB!

src/trace/context/extractors/http.ts
```diff
@@ -41,11 +41,17 @@ export class HTTPEventTraceExtractor implements EventTraceExtractor {
...
+        // Single‐value header
+        lowerCaseHeaders[key.toLowerCase()] = val;
+      }
```
duncanista, line 54:
```suggestion
      if (Array.isArray(val)) {
        // MultiValueHeaders: take the first value
        lowerCaseHeaders[key.toLowerCase()] = val[0] ?? "";
        continue;
      }
      
      // Single‐value header
      lowerCaseHeaders[key.toLowerCase()] = val;
      }
```

duncanista (approved):

PR #648, opened 2025-05-07
Implement dogstatsd, add timestamp support, fix flushing

nhulston:
### What does this PR do?

- Removes `hot-shots`
- Implements dogstatsd ourselves
- Updates dogstatsd implementation to support timestamps
- Therefore, if the user has extension enabled and sends a metric with a timestamp, we can now use dogstatsd instead of directly using the API
- This PR also implements a working `flush`. The previous implementation just closed the UDP socket on "flush" which actually just drops any unsent UDP packets, leading to lost metrics.
- Just use one socket, no need to keep opening/closing sockets on every invocation

### Motivation

- `hot-shots` did not support metrics with timestamps, so originally we just sent metrics directly using the Datadog API
- However, the DD API is not FIPs compliant. This seemed like the easiest solution
- dogstatsd is a very simple protocol so might as well do it ourselves. Now, our performance should be slightly faster and our package size should be slightly smaller!

### Testing Guidelines

- Tested manually with a Lambda, calling `sendDistributionMetric` and `sendDistributionMetricWithDate`. Both work!
- Unit tests

### Additional Notes

- This should be non-breaking, but this is a big change, so we should be careful and test a lot with self-monitoring before the next release.
- Will deal with FIPs compliance stuff in a separate PR after this

### Types of Changes

- [x] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -509,7 +509,7 @@ describe("datadog", () => {
 
     expect(mockedIncrementInvocations).toBeCalledTimes(1);
     expect(mockedIncrementInvocations).toBeCalledWith(expect.anything(), mockContext);
-    expect(logger.debug).toHaveBeenCalledTimes(8);
+    expect(logger.debug).toHaveBeenCalledTimes(9);
     expect(logger.debug).toHaveBeenLastCalledWith('{"status":"debug","message":"datadog:Unpatching HTTP libraries"}');
   });
 
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -509,7 +509,6 @@ describe("datadog", () => {
 
     expect(mockedIncrementInvocations).toBeCalledTimes(1);
     expect(mockedIncrementInvocations).toBeCalledWith(expect.anything(), mockContext);
-    expect(logger.debug).toHaveBeenCalledTimes(9);
     expect(logger.debug).toHaveBeenLastCalledWith('{"status":"debug","message":"datadog:Unpatching HTTP libraries"}');
   });
 
--- a/src/metrics/dogstatsd.ts
+++ b/src/metrics/dogstatsd.ts
@@ -40,7 +40,7 @@ export class LambdaDogStatsD {
    * Send a distribution value, optionally setting tags and timestamp.
    * Timestamp is seconds since epoch.
    */
-  public distribution(metric: string, value: number, tags?: string[], timestamp?: number): void {
+  public distribution(metric: string, value: number, timestamp?: number, tags?: string[]): void {
     this.report(metric, "d", value, tags, timestamp);
   }
 
--- a/src/metrics/dogstatsd.ts
+++ b/src/metrics/dogstatsd.ts
@@ -36,7 +36,7 @@ export class LambdaDogStatsD {
         logDebug(`Socket send buffer increased to ${LambdaDogStatsD.MIN_SEND_BUFFER_SIZE / 1024}kb`);
       }
     } catch {
-      logDebug("Unable to set socket's send buffer size")
+      logDebug("Unable to set socket's send buffer size");
     }
   }
 
--- a/src/metrics/dogstatsd.ts
+++ b/src/metrics/dogstatsd.ts
@@ -1,0 +1,38 @@
+import * as dgram from "node:dgram";
+import { SocketType } from "node:dgram";
+
+export class LambdaDogStatsD {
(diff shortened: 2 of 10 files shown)


src/metrics/dogstatsd.ts
```diff
@@ -0,0 +1,92 @@
...
+  private normalizeTags(tags: string[]): string[] {
+    return tags.map((t) => t.replace(LambdaDogStatsD.TAG_RE, LambdaDogStatsD.TAG_SUB));
+  }
+
+  private report(metric: string, metricType: string, value: number | null, tags?: string[], timestamp?: number): void {
+    if (value == null) {
+      return;
+    }
+    const serializedTags = tags && tags.length ? `|#${this.normalizeTags(tags).join(",")}` : "";
+    const timeStampPart = timestamp != null ? `|T${timestamp}` : "";
```
duncanista, line 57:
```suggestion
    const timestampPart = timestamp != null ? `|T${timestamp}` : "";
```

src/metrics/listener.ts
```diff
@@ -146,33 +134,30 @@ export class MetricsListener {
...
-        }
-      } else {
-        this.statsDClient?.distribution(name, value, undefined, tags);
+      const dateCeiling = new Date(Date.now() - HISTORICAL_METRICS_THRESHOLD_HOURS); // 4 hours ago
+      if (dateCeiling > metricTime) {
+        logWarning(`Timestamp ${metricTime.toISOString()} is older than 4 hours, not submitting metric ${name}`);
         return;
       }
+
+      const secondsSinceEpoch = Math.floor(metricTime.getTime() / 1000);
```
duncanista, line 145: Don't we have millisecond/nanosecond resolution?

src/metrics/listener.ts
```diff
@@ -146,33 +134,30 @@ export class MetricsListener {
...
+      this.statsDClient?.distribution(name, value, secondsSinceEpoch, tags);
```
duncanista, line 146:
```suggestion
      this.statsDClient.distribution(name, value, secondsSinceEpoch, tags);
```

duncanista (commented):
Left some comments – would like to see if we can do millisecond/nanosecond resolution in Extension/Forwarder, just a nit

src/metrics/listener.ts
nhulston: The unit passed in must be seconds, but it looks like it can take decimals, so MS resolution works!

nhulston:
> Left some comments – would like to see if we can do millisecond/nanosecond resolution in Extension/Forwarder, just a nit

We discussed in Slack but replying here for visibility:
- The extension rounds metrics to the nearest 10 second bucket, so rounding here is fine
- Also, timestamps with decimals are currently broken (this is a regression), but I opened a fix here: https://github.com/DataDog/serverless-components/pull/17
- In case customers in the future use old versions of the extension and new versions of datadog-lambda-js, I will make sure to round here for backwards compatibility

src/metrics/dogstatsd.ts
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    try {
+      const currentSize = sock.getSendBufferSize();
+      if (currentSize <= LambdaDogStatsD.MIN_SEND_BUFFER_SIZE) {
+        sock.setSendBufferSize(LambdaDogStatsD.MIN_SEND_BUFFER_SIZE);
+        logDebug(`Socket send buffer increased to ${LambdaDogStatsD.MIN_SEND_BUFFER_SIZE / 1024}kb`);
+      }
+    } catch {
+      // ignore
```
duncanista, line 36: why?

duncanista (approved):

src/metrics/dogstatsd.ts
nhulston: Added a debug log

src/index.spec.ts
```diff
@@ -509,7 +509,7 @@ describe("datadog", () => {
 
     expect(mockedIncrementInvocations).toBeCalledTimes(1);
     expect(mockedIncrementInvocations).toBeCalledWith(expect.anything(), mockContext);
-    expect(logger.debug).toHaveBeenCalledTimes(8);
+    expect(logger.debug).toHaveBeenCalledTimes(9);
```
nhulston, line 512: we get another debug log for increasing buffer send size

PR #658, opened 2025-06-03
chore: Use GitHub App for update-deps workflow

lym953:
### What does this PR do?

Makes the upgrade workflow use GitHub Apps to get temporary tokens instead of using a long-lived SSH key.

### Motivation

This workflow is broken, probably because the SSH key no longer works. Instead of creating a new SSH key, we use a GitHub App, which generates short-lived tokens and is more secure than SSH key.

### Testing Guidelines

Will test the workflow after this PR is merged to `main` because I can't manually trigger it right now.

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

Jira: https://datadoghq.atlassian.net/browse/SVLS-6929

### DIFF
--- a/.github/workflows/update-deps.yml
+++ b/.github/workflows/update-deps.yml
@@ -3,6 +3,7 @@ name: update-deps
 on:
   schedule:
     - cron: "0 10 * * *" # Run at 10 am every day
+  workflow_dispatch:
 
 jobs:
   check:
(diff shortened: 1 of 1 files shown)


.github/workflows/update-deps.yml
```diff
@@ -3,14 +3,24 @@ name: update-deps
 on:
   schedule:
     - cron: "0 10 * * *" # Run at 10 am every day
+  workflow_dispatch:
```
lym953, line 6: Just adding it so I can manually trigger it to test it.

alexgallotta (approved):

PR #666, opened 2025-07-12
Remove Service Override from `aws.lambda` spans

zarirhamza:
### What does this PR do?
Change the lambda span service name priority:
1) value of `DD_SERVICE`
2) value of lambda function name
3) `aws.lambda`

### Motivation

Removes service overrides for `aws.lambda` span

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/trace/listener.ts
+++ b/src/trace/listener.ts
@@ -7,7 +7,7 @@ import { ColdStartTracerConfig, ColdStartTracer } from "./cold-start-tracer";
 import { logDebug, tagObject } from "../utils";
 import { didFunctionColdStart, isProactiveInitialization } from "../utils/cold-start";
 import { datadogLambdaVersion } from "../constants";
-import { ddtraceVersion, parentSpanFinishTimeHeader } from "./constants";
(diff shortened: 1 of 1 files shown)


astuyve (changes requested):
I don't think this can ship. Many assumptions are made in the backend based on this

zarirhamza:
> I don't think this can ship. Many assumptions are made in the backend based on this

What part exactly is blocking? Is it changing the service name from `aws.lambda` to functionName? That change was already merged in #655. The only thing to look out for was span.name according to Rey and that is different from service. span.name is akin to operation name which is unchanged with these changes.

Or is it respecting the DD_SERVICE env var specifically for lambda spans? That isn't really changing any behavior because the rust [extension](https://github.com/DataDog/datadog-lambda-extension/blob/e3a9dc43297ecb43a2581d1975afb5b1c388ae39/bottlecap/src/traces/trace_processor.rs#L47) already does that. This change simply performs that renaming earlier in the life of the trace so that base_service is not on the tracer leading to indications of a service override on the UI even though the base_service and actual service are the same.

zarirhamza:
Reminder - Use this PR to collapse cold start to use Boolean values

PR #672, opened 2025-07-28
feat: lambda support for DSM 

michael-zhao459:
### What does this PR do?
This PR adds lambda support for Data Streams Monitoring (DSM), sets a checkpoint after extracting context from trace propagation headers, DSM context gets packaged in with the trace prop headers. Queues supported by DSM include SQS, SNS, SNS->SQS, and Kinesis

### Motivation

Consume calls of lambdas are set as triggers to queue and do not have any explicit calls the tracers can hook into to set a checkpoint

### Testing Guidelines

Tried all configurations with sandbox AWS account. Wrote unit tests ensuring all cases were covered. 

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


src/trace/context/extractors/kinesis.ts
```diff
@@ -8,15 +8,19 @@ export class KinesisEventTraceExtractor implements EventTraceExtractor {
...
 
+    sourceARN = event.Records[0].eventSourceARN;
+
     try {
       const decodedData = Buffer.from(kinesisData, "base64").toString("ascii");
       const parsedBody = JSON.parse(decodedData);
       const headers = parsedBody?._datadog;
       if (headers) {
         const traceContext = this.tracerWrapper.extract(headers);
+        this.tracerWrapper.setConsumeCheckpoint(headers, "kinesis", sourceARN);
```
michael-zhao459, line 23: Need the extra parameter will add later, I just like seeing tests pass

src/trace/context/extractors/kinesis.spec.ts
```diff
@@ -67,25 +80,50 @@ describe("KinesisEventTraceExtractor", () => {
...
       const traceContext = extractor.extract(payload as any);
       expect(traceContext).toBeNull();
+
+      expect(mockDataStreamsCheckpointer.setConsumeCheckpoint).toHaveBeenCalledTimes(dsmCalls);
+
+      if (dsmCalls > 0) {
+        expect(mockDataStreamsCheckpointer.setConsumeCheckpoint).toHaveBeenCalledWith(
+          "kinesis",
+          "arn:aws:kinesis:test",
+          null,
```
michael-zhao459, line 123: The param I'm actually testing for

src/trace/context/extractors/kinesis.ts
```diff
@@ -8,15 +8,19 @@ export class KinesisEventTraceExtractor implements EventTraceExtractor {
   constructor(private tracerWrapper: TracerWrapper) {}
 
   extract(event: KinesisStreamEvent): SpanContextWrapper | null {
+    let sourceARN = "";
     const kinesisData = event?.Records?.[0]?.kinesis.data;
     if (kinesisData === undefined) return null;
 
+    sourceARN = event.Records[0].eventSourceARN;
```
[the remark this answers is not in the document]
michael-zhao459: The consume call of the AWS queue type can be done in a batch for SQS and Kinesis (SNS is one at a time, still in Records array for some reason), this is not Lambda specific. For the tracers, when the consuming service calls sqs.receiveMessage, we only set one checkpoint for that invocation despite it possibly batching messages. This is consistent with what we are doing here. This is the same implementation as the python Lambda code as well.

src/trace/context/extractors/kinesis.ts
```diff
@@ -8,26 +8,31 @@ export class KinesisEventTraceExtractor implements EventTraceExtractor {
...
+        this.tracerWrapper.setConsumeCheckpoint(headers, "kinesis", record.eventSourceARN);
 
-        logDebug(`Extracted trace context from Kinesis event`, { traceContext, headers });
-        return traceContext;
-      }
-    } catch (error) {
-      if (error instanceof Error) {
-        logDebug("Unable to extract trace context from Kinesis event", error);
+        // If we already have a context, we can skip the rest of the records
+        if (context) continue;
```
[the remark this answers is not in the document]
ericfirth: done

purple4reina (approved):
We should break after the first loop in all these situations when dsm is disabled.

src/trace/context/extractors/kinesis.ts
```diff
@@ -8,26 +8,35 @@ export class KinesisEventTraceExtractor implements EventTraceExtractor {
...
-        return traceContext;
-      }
-    } catch (error) {
-      if (error instanceof Error) {
-        logDebug("Unable to extract trace context from Kinesis event", error);
+        // If we already have a context, we can skip the rest of the records
+        // also, if DSM is disabled, we can stop extracting context after the first record
+        if (!this.tracerWrapper.isDataStreamsEnabled && context) {
+          break;
+        }
```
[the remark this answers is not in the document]
purple4reina: In fact, I think we should be breaking every time when dsm is not enabled. That is what the code was doing before this PR, so we should maintain that functionality.

src/trace/tracer-wrapper.ts
```diff
@@ -99,4 +100,28 @@ export class TracerWrapper {
...
+      if (err instanceof Object || err instanceof Error) {
+        logDebug(`DSM: Failed to set consume checkpoint for ${eventType} ${arn}:`, err);
+      }
+    }
+  }
+
+  public get isDataStreamsEnabled(): boolean {
+    const validEnabledValues = new Set(["true", "1", "yes", "y", "on", "1"]);
+    return validEnabledValues.has(getEnvValue("DD_DATA_STREAMS_ENABLED", "false").toLowerCase());
+  }
```
purple4reina, line 126: We should move this config check to the index.ts file, so it sits along with all the other configuration and means we only need to read the env var once.

src/trace/context/extractors/sns-sqs.ts
```diff
@@ -10,50 +10,63 @@ export class SNSSQSEventTraceExtractor implements EventTraceExtractor {
...
-        if (traceContext) {
-          return traceContext;
+        // Set a checkpoint for the record, even if we don't have headers
+        this.tracerWrapper.setConsumeCheckpoint(headers, "sqs", record.eventSourceARN);
+
+        // If we've already extracted context, skip the rest of the extraction, since we only want to extract context once
+        // also, if DSM is disabled, we can break out of the loop early
+        if (!this.tracerWrapper.isDataStreamsEnabled && context) {
+          break;
         }
```
purple4reina, line 49: Again, I think this break should be moved directly after where the context object is defined.

purple4reina: Actually. I think we need to go one step further. Even if there is no context found, we should be returning after the first loop when dsm is disabled.

ericfirth: done

src/index.ts
```diff
@@ -52,6 +52,15 @@ export const minColdStartTraceDurationEnvVar = "DD_MIN_COLD_START_DURATION";
...
 export const addSpanPointersEnvVar = "DD_TRACE_AWS_ADD_SPAN_POINTERS";
+export const dataStreamsEnabledEnvVar = "DD_DATA_STREAMS_ENABLED";
+
+/**
+ * Check if Data Streams Monitoring is enabled
+ * Reads from environment variable through config system
+ */
+export function getDataStreamsEnabled(): boolean {
+  return getConfig().dataStreamsEnabled;
+}
```
purple4reina, line 63: This is not what we want. Calling `getConfig` will read each and every environment variable each and every time it is called.

Instead, you will want to use the global config that is created just once at start up.

Pass `this.config` to the constructor of the trace extractor https://github.com/DataDog/datadog-lambda-js/blob/1619bce8fa17d7c0e64044e471161e5108e1713d/src/trace/context/extractor.ts#L89

then you can use it during extraction without having to read environment variables each time.

purple4reina (changes requested):
Two requested changes:
1. Read from global config object
2. Always break from loop when dsm not enabled

src/index.ts
ericfirth: thanks for clarifying. i am not super-familiar with this repo. made this change as well

src/trace/tracer-wrapper.ts
```diff
@@ -99,4 +116,23 @@ export class TracerWrapper {
...
+
+  public setConsumeCheckpoint(contextJson: any, eventType: string, arn: string): void {
+    if (!arn) {
+      logDebug("DSM: No ARN provided, skipping setConsumeCheckpoint");
+      return;
+    }
+
+    if (!this.config.dataStreamsEnabled) {
+      return;
+    }
```
purple4reina, line 128: Can we put this before the arn check, that way when dsm is disabled, users will never see dsm related messages in their logs.

src/index.ts
```diff
@@ -465,16 +473,16 @@ export async function emitTelemetryOnErrorOutsideHandler(
       },
       startTime,
     };
-    const tracerWrapper = new TracerWrapper();
+    const config = getConfig();
+    const tracerWrapper = new TracerWrapper(config);
```
purple4reina, line 477: I don't understand why is this change necessary? This method is only called when there is an error attempting to initialize the handler.

src/trace/tracer-wrapper.ts
```diff
@@ -27,8 +28,24 @@ export interface TraceOptions {
...
+      createInferredSpan: true,
+      encodeAuthorizerContext: true,
+      decodeAuthorizerContext: true,
+      mergeDatadogXrayTraces: false,
+      injectLogContext: false,
+      minColdStartTraceDuration: 3,
+      coldStartTraceSkipLib: "",
+      addSpanPointers: true,
+      dataStreamsEnabled: false,
+    };
```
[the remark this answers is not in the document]
purple4reina: In fact, we don't need to add the config to this class in the first place. We should just rely on the config on the extractor alone.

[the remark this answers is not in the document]
ericfirth: 👍 true. I think I was being too deferential to the change Michael made where there was a check in the tracer wrapper because set consume checkpoint was always run. But with the optimizations where we are only setting checkpoints when dsm is enabled, so there is no need for the second check

src/trace/tracer-wrapper.ts
```diff
@@ -99,4 +99,19 @@ export class TracerWrapper {
...
   }
+
+  public setConsumeCheckpoint(contextJson: any, eventType: string, arn: string): void {
+    if (!arn) {
+      logDebug("DSM: No ARN provided, skipping setConsumeCheckpoint");
+      return;
+    }
+
+    try {
+      this.tracer.dataStreamsCheckpointer.setConsumeCheckpoint(eventType, arn, contextJson, false);
```
purple4reina, line 110: Question: is there a particular version of the tracer that this requires?

src/trace/context/extractors/sqs.ts
```diff
@@ -1,33 +1,45 @@
...
+    if (this.config.dataStreamsEnabled) {
+      for (let i = 0; i < (event?.Records || []).length; i++) {
+        const record = event.Records[i];
+        try {
+          const headers = this.getParsedRecordHeaders(record);
 
-      if (!headers) {
-        // Then try to get from binary value. This happens when SNS->SQS, but SNS has raw message delivery enabled.
-        // In this case, SNS maps any messageAttributes to the SQS messageAttributes.
-        // We can at least get trace context from SQS, but we won't be able to create the SNS inferred span.
```
purple4reina, line 20: Can we keep this comment please?

src/trace/tracer-wrapper.ts
ericfirth: meaning the 4th argument (whether its a manually instrumented or not)? that is on 5.62.0, https://github.com/DataDog/dd-trace-js/releases/tag/v5.62.0

purple4reina: In that case, you will need to bump the dependency version of ddtrace. https://github.com/DataDog/datadog-lambda-js/blob/1619bce8fa17d7c0e64044e471161e5108e1713d/package.json#L32

src/trace/context/extractors/kinesis.spec.ts
```diff
@@ -112,10 +260,78 @@ describe("KinesisEventTraceExtractor", () => {
...
       };
 
-      const extractor = new KinesisEventTraceExtractor(tracerWrapper);
+      const extractor = new KinesisEventTraceExtractor(tracerWrapper, mockConfig);
 
       const traceContext = extractor.extract(payload);
       expect(traceContext).toBeNull();
     });
+
+    it("extracts trace context from multiple records when DSM is disabled but does not call setConsumeCheckpoint", () => {
```
purple4reina, line 269: This seems to be wrong now, we only extract from the first record, yeah?

PR #674, opened 2025-08-14
feat: add configurable support for span links instead of parent-child

jeastham1993:
### What does this PR do?

Adds an optional configuration option for using span links instead of the automatic parenting to the upstream trace context.

### Motivation

Parent-child relationships are not always optimal when building asynchronous systems, OpenTelemetry actually recommedn using span links in most cases. Adding a configurable option allows developers to pick, for each Lambda function, if parenting or linking is the right approach.

### Testing Guidelines

Added tests, need advice on best ways to test this with an actual function.

### Types of Changes

- [ ] Bug fix
- [ X ] New feature
- [ ] Breaking change

### Check all that apply

- [ X ] This PR's description is comprehensive
- [ X ] This PR introduces new APIs or parameters that are documented and unlikely to change in the foreseeable future
- [ X ] This PR impacts documentation, and it has been updated (or a ticket has been logged)
- [ X ] This PR's changes are covered by the automated tests

### DIFF
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -49,19 +49,19 @@ let mockTraceSource: TraceSource | undefined = undefined;
 
 jest.mock("./trace/trace-context-service", () => {
   class MockTraceContextService {
-    extract(event: any, context: Context): SpanContextWrapper {
-      return mockSpanContextWrapper;
+    extract(event: any, context: Context): SpanContextWrapper[] {
+      return [mockSpanContextWrapper];
     }
 
     get traceSource() {
       return mockTraceSource;
     }
     get currentTraceContext() {
-      return mockSpanContextWrapper;
+      return [mockSpanContextWrapper];
     }
 
     get currentTraceHeaders() {
-      return mockTraceHeaders;
+      return [mockTraceHeaders];
     }
   }
   return {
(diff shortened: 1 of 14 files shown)


nhulston:
This looks really good! I haven't tested that it works yet, but I'm happy to do so.

Currently, this would only create a span link for the first item in a batch, but span links are ideal for a span with multiple 'parents'. Here's the steps I would take to allow for multiple span links:

1. [`extractor.ts`] Update EventTraceExtractor interface and TraceContextExtractor.extract() to return arrays instead of a single trace context
2. [`sqs.ts/sns.ts/event-bridge.ts/anything else I missed`] Update extract() to return arrays. If span links are enabled, loop through all `event.Records` instead of just accessing `event.Record[0]`. We can limit this to a maximum of, for example, 20 records, so we don't waste time and overwhelm the user with too much information. If span links are not enabled, keep accessing `event.Record[0]` and just return that as a size-1 array.
3. [`trace-context-service.ts` and `listener.ts`] Handle arrays instead of a single value
4. [`listener.ts`][onWrap()] Create span links for each item in the array

jeastham1993:
@nhulston I've finally got around to adding the additional changes to this PR. I'll fix the merge conflicts, but any chance you could have another look over this now that I've added more robust support for span links?

All the unit tests pass. I'm just deploying stuff to my AWS account to test it works when deployed. I might not get around to doing that before I go on PTO so wanted to leave this for you to have a look at.

PR #676, opened 2025-09-23
fix: reset tracecontext to avoid unintentional caching

joeyzhao2018:
### What does this PR do?
The customer is experiencing a case where multiple lambda invocations are put into one trace even though they have different tracecontext in their payload. This indicates that the `rootTraceContext` was reused/cached unintentionally. 

### Motivation
https://datadoghq.atlassian.net/browse/APMS-17080

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -63,6 +63,10 @@ jest.mock("./trace/trace-context-service", () => {
     get currentTraceHeaders() {
       return mockTraceHeaders;
     }
+
+    reset(){
+      // mocking
+    }
   }
   return {
     ...jest.requireActual("./trace/trace-context-service"),
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -64,7 +64,7 @@ jest.mock("./trace/trace-context-service", () => {
       return mockTraceHeaders;
     }
 
-    reset(){
+    reset() {
       // mocking
     }
   }
--- a/src/trace/listener.spec.ts
+++ b/src/trace/listener.spec.ts
@@ -68,6 +68,8 @@ jest.mock("./trace-context-service", () => {
     get currentTraceContext() {
       return mockSpanContextWrapper;
     }
+    reset() { // mocking
+    }
   }
   return {
     ...jest.requireActual("./trace-context-service"),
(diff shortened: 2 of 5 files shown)


lucaspimentel:
@codex review

lucaspimentel (approved):

joeyzhao2018:
[HOLD] Not merging yet because this seems to be a broader issue if the fix is correct. So we should reproduce first.
Asked AI about what kind of payloads can cause this and the answer is as following

```
The question is about when rootTraceContext gets reused after extraction failure. Looking at the recent commit, there was
  a bug where rootTraceContext wasn't being reset between invocations, causing unintentional caching.
Based on the extraction flow, the rootTraceContext would be reused (before the recent fix) in these scenarios:

  Event payloads that would reach XRay fallback extraction:

  1. Plain object events without recognized structure - Events that don't match any of the supported event types (HTTP, SNS,
   SQS, Kinesis, EventBridge, etc.) and have no custom extractor configured.
  2. Events where all extraction methods fail - For example:
    - HTTP events with malformed or missing trace headers
    - SNS/SQS events without proper trace context in message attributes
    - Kinesis events with corrupted or missing trace data
    - Events where the Lambda context itself has no X-Ray trace information
  3. Malformed events - Events that are not objects or have unexpected structure that causes extractors to return null.

  Example event payloads that would reach XRay extraction:
  // Plain custom event with no trace headers
  { "customData": "value", "timestamp": 123456789 }

  // HTTP event with missing/invalid trace headers  
  { "headers": {}, "body": "..." }
```
Trying these now.
  // SNS event with no trace context in message attributes
  { "Records": [{ "Sns": { "Message": "...", "MessageAttributes": {} } }] }

  The recent fix ensures rootTraceContext = null at the start of each extraction, preventing stale trace context from
  previous invocations being reused.

joeyzhao2018:
I could not reproduce. We actually have the exact payload in this customer's case but somehow it's not getting any failure yet the extracted context is different from the customer's case, which is different from any of the tracecontext propagated.

PR #744, opened 2026-03-16
feat(aap): In App WAF support

CarlesDD:
### What does this PR do?

Adds AppSec support to the Lambda layer by extracting HTTP data from Lambda events and publishing it to diagnostic channels consumed by dd-trace-js's AppSec subsystem.

inappwafport

Changes:
- **Dockerfile + move script**: `move_ddtrace_dependency.js` now reads `@datadog/native-appsec` from dd-trace's `optionalDependencies` and promotes it to a direct dependency so it survives `--ignore-optional`. The Dockerfile runs the script before `rm -rf node_modules` and strips unused native prebuilds (non-Linux-glibc platforms)
- **Event data extractor** (`src/appsec/event-data-extractor.ts`): Parses API Gateway v1/v2, ALB, and Lambda Function URL events, extracting headers, method, path, query, body, client IP, path params, cookies, and route
- **Orchestrator** (`src/appsec/index.ts`): Checks `DD_APPSEC_ENABLED` and publishes extracted data to `datadog:lambda:start-invocation` / `datadog:lambda:end-invocation` diagnostic channels
- **TraceListener integration**: `initAppsec()` called in `onStartInvocation`, `processAppSecRequest` and `processAppSecResponse` called in `onEndingInvocation`

### Motivation

Porting the In-App WAF security product to AWS Lambda for the Node.js runtime. The Lambda layer extracts HTTP data and dispatches it to the tracer for WAF execution and reporting.

The layer is intentionally kept thin, only extracting and publishing data. All security logic (WAF, reporting, trace keeping) lives in dd-trace-js.

### Testing Guidelines

- Unit tests for the event data extractor cover all 4 HTTP event types (API GW v1/v2, ALB, Lambda Function URL) and non-HTTP events
- Unit tests for the orchestrator verify configuration gating and channel publishing
- Existing listener tests pass unchanged

### Additional Notes

- `@datadog/native-appsec` is NOT added to this repo's `package.json`. The version is read dynamically from dd-trace's `optionalDependencies` at build time, so dd-trace-js remains the single owner of the native module version.
- No new environment variables, using the existing `DD_APPSEC_ENABLED`.
- This is a monitoring-only first iteration. Blocking, Remote Config, and telemetry are out of scope.
- Companion PR in dd-trace-js: https://github.com/DataDog/dd-trace-js/pull/7783

### Types of Changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

- [x] This PR introduces new APIs or parameters that are documented and unlikely to change in the foreseeable future

[APPSEC-60752](https://datadoghq.atlassian.net/browse/APPSEC-60752)

[APPSEC-60752]: https://datadoghq.atlassian.net/browse/APPSEC-60752?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/move_ddtrace_dependency.js
+++ b/scripts/move_ddtrace_dependency.js
@@ -1,4 +1,6 @@
 // Moves the dd-trace dependency from devDependencies to dependencies within package.json.
+// Also promotes selected dd-trace optionalDependencies to direct dependencies so they
+// survive `yarn install --production=true --ignore-optional`.
 // This is used when building the Layer
 
 // USAGE: ./move_dd_trace_dependency.js "$(cat package.json)" > package.json
@@ -10,6 +12,8 @@ moveDependency('@datadog/pprof')
 moveDependency('@opentelemetry/api')
 moveDependency('@opentelemetry/api-logs')
 
+addOptionalFromDdTrace('@datadog/native-appsec')
+
 console.log(JSON.stringify(file, null, 2));
 
 function moveDependency (name) {
@@ -17,3 +21,16 @@ function moveDependency (name) {
   delete file.devDependencies[name];
   file.dependencies[name] = ddTraceVersion;
 }
+
+function addOptionalFromDdTrace (name) {
+  try {
+    const ddTracePkg = require('dd-trace/package.json')
+    const version = ddTracePkg.optionalDependencies?.[name]
+    if (version) {
+      file.dependencies[name] = version
+    }
+  } catch {
+    // dd-trace not installed yet; skip
+  }
+}
+
--- a/scripts/move_ddtrace_dependency.js
+++ b/scripts/move_ddtrace_dependency.js
@@ -1,6 +1,4 @@
 // Moves the dd-trace dependency from devDependencies to dependencies within package.json.
-// Also promotes selected dd-trace optionalDependencies to direct dependencies so they
-// survive `yarn install --production=true --ignore-optional`.
 // This is used when building the Layer
 
 // USAGE: ./move_dd_trace_dependency.js "$(cat package.json)" > package.json
@@ -20,4 +18,3 @@ function moveDependency (name) {
   delete file.devDependencies[name];
   file.dependencies[name] = ddTraceVersion;
 }
-
--- a/scripts/move_ddtrace_dependency.js
+++ b/scripts/move_ddtrace_dependency.js
@@ -8,11 +8,10 @@
 const file = JSON.parse(process.argv[2]);
 
 moveDependency('dd-trace')
+moveDependency('@datadog/native-appsec')
 moveDependency('@datadog/pprof')
 moveDependency('@opentelemetry/api')
 moveDependency('@opentelemetry/api-logs')
-
-addOptionalFromDdTrace('@datadog/native-appsec')
 
 console.log(JSON.stringify(file, null, 2));
 
@@ -20,17 +19,5 @@ function moveDependency (name) {
   const ddTraceVersion = file.devDependencies[name];
   delete file.devDependencies[name];
   file.dependencies[name] = ddTraceVersion;
-}
-
-function addOptionalFromDdTrace (name) {
-  try {
-    const ddTracePkg = require('dd-trace/package.json')
-    const version = ddTracePkg.optionalDependencies?.[name]
-    if (version) {
-      file.dependencies[name] = version
-    }
-  } catch {
-    // dd-trace not installed yet; skip
-  }
 }
 
--- a/src/appsec/event-data-extractor.ts
+++ b/src/appsec/event-data-extractor.ts
@@ -1,0 +1,229 @@
+import * as eventType from "../utils/event-type-guards";
+
+export interface ExtractedHTTPData {
+  headers: Record<string, string>;
+  method: string;
+  path: string;
+  query?: Record<string, string | string[]>;
+  body?: string | object;
+  isBase64Encoded: boolean;
+  clientIp?: string;
+  pathParams?: Record<string, string>;
+  cookies?: Record<string, string>;
+  route?: string;
+}
+
+export function extractHTTPDataFromEvent(event: any): ExtractedHTTPData | undefined {
+  if (eventType.isLambdaUrlEvent(event)) {
+    return extractFromLambdaUrl(event);
+  }
+
+  if (eventType.isAPIGatewayEvent(event)) {
+    return extractFromApiGatewayV1(event);
+  }
+
+  if (eventType.isAPIGatewayEventV2(event)) {
+    return extractFromApiGatewayV2(event);
+  }
+
+  if (eventType.isALBEvent(event)) {
+    return extractFromALB(event);
+  }
+
+  return undefined;
+}
+
+function extractFromApiGatewayV1(event: any): ExtractedHTTPData {
+  const headers = normalizeHeaders(event.headers, event.multiValueHeaders);
+  const { cookies, headersNoCookies } = separateCookies(headers);
+
+  return {
+    headers: headersNoCookies,
+    method: event.httpMethod || "",
+    path: event.requestContext?.path || event.path || "/",
+    query: mergeQueryParams(event.queryStringParameters, event.multiValueQueryStringParameters),
+    body: decodeBody(event.body, event.isBase64Encoded),
+    isBase64Encoded: !!event.isBase64Encoded,
+    clientIp: event.requestContext?.identity?.sourceIp,
+    pathParams: event.pathParameters || undefined,
+    cookies,
+    route: event.resource,
+  };
+}
+
+function extractFromApiGatewayV2(event: any): ExtractedHTTPData {
+  const headers = normalizeHeaders(event.headers);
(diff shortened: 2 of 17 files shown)


src/trace/listener.ts
```diff
@@ -175,6 +179,9 @@ export class TraceListener {
     // so we won't crash user code.
     if (!this.tracerWrapper.currentSpan) return false;
     this.wrappedCurrentSpan = new SpanWrapper(this.tracerWrapper.currentSpan, {});
+
+    processAppsecResponse(event, this.tracerWrapper.currentSpan);
```
IlyasShabi: The signature of this function is `span: any, statusCode?: string, responseHeaders?: Record<string, string>` wondering if this correct way to call it

src/appsec/event-data-extractor.ts
```diff
@@ -0,0 +1,229 @@
...
+}
+
+function separateCookies(headers: Record<string, string>): {
+  cookies: Record<string, string> | undefined;
+  headersNoCookies: Record<string, string>;
+} {
+  const cookies = parseCookieHeader(headers.cookie);
+  const headersNoCookies = { ...headers };
+  delete headersNoCookies.cookie;
+  return { cookies, headersNoCookies };
```
IlyasShabi: If cookies is `undefined` returned from `parseCookieHeader` we should not add it right?

scripts/move_ddtrace_dependency.js
```diff
@@ -10,10 +12,25 @@ moveDependency('@datadog/pprof')
 moveDependency('@opentelemetry/api')
 moveDependency('@opentelemetry/api-logs')
 
+addOptionalFromDdTrace('@datadog/native-appsec')
```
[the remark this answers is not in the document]
CarlesDD: I've changed the implementation to mimic datadog/pprof package using `moveDependency`

src/trace/listener.ts
CarlesDD: Good catch. Fixed and included a test

src/appsec/event-data-extractor.ts
CarlesDD: You're right

src/appsec/event-data-extractor.ts
```diff
@@ -0,0 +1,229 @@
...
+  };
+}
+
+function extractFromApiGatewayV2(event: any): ExtractedHTTPData {
+  const headers = normalizeHeaders(event.headers);
+  const { headersNoCookies } = separateCookies(headers);
+
+  const cookies = parseCookieArray(event.cookies) || parseCookieHeader(headers.cookie);
+
+  let route: string | undefined;
```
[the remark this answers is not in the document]
CarlesDD: It could be for v2. Fixed!

src/appsec/index.ts
```diff
@@ -0,0 +1,47 @@
...
+import { extractHTTPDataFromEvent } from "./event-data-extractor";
+
+const startInvocationChannel = dc.channel("datadog:lambda:start-invocation");
+const endInvocationChannel = dc.channel("datadog:lambda:end-invocation");
+
+let enabled = false;
+
+export function initAppsec(): void {
+  const envValue = process.env.DD_APPSEC_ENABLED;
+  enabled = envValue === "true" || envValue === "1";
```
uurien: dd-trace-js is case insensitive for boolean config values, I think that we should do the same here:

https://github.com/DataDog/dd-trace-js/blob/6d1e0726976dd28073402d85ff89d7bd55031d94/packages/dd-trace/src/config/parsers.js#L192-L203

IlyasShabi (approved):

src/trace/listener.ts
```diff
@@ -179,6 +183,9 @@ export class TraceListener {
     // so we won't crash user code.
     if (!this.tracerWrapper.currentSpan) return false;
     this.wrappedCurrentSpan = new SpanWrapper(this.tracerWrapper.currentSpan, {});
+
+    processAppsecRequest(event, this.tracerWrapper.currentSpan);
```
purple4reina: Should this happen at the beginning of the invocation? I'm wondering if this should be moved to `onStartInvocation`.

CarlesDD: Good catch. `processAppsecRequest` ran after the handler, too late to ever support blocking. Moved it to `onRequestStart`, called right before the handler runs, once the span is active. Couldnt be `onStartInvocation` itself since the span doesnt exist yet there.

src/trace/listener.ts
```diff
@@ -164,6 +168,16 @@ export class TraceListener {
...
   }
 
+  /**
+   * onRequestStart runs once the aws.lambda span is active, before the user
+   * function is invoked.
+   *
+   * @param event
+   */
+  public onRequestStart(event: any): void {
+    processAppsecRequest(event, this.tracerWrapper.currentSpan);
```
purple4reina, line 182: Is there a reason we can't just put this into `onStartInvocation`?

src/trace/listener.ts
```diff
@@ -115,6 +117,8 @@ export class TraceListener {
   }
 
   public async onStartInvocation(event: any, context: Context) {
+    initAppsec();
```
purple4reina: Looks like there's a `this.config` object available. Makes sense to add the appsec enabled check to the config.

src/trace/listener.ts
CarlesDD: Yes, there is. `onStartInvocation` runs before `onWrap` even creates the span (`tracer.wrap()` activates it later, wrapping the handler invocation itself), so `currentSpan` would be `null` there. `onRequestStart` is the earliest point where the span is guaranteed active, right before the handler runs.

src/trace/listener.ts
CarlesDD: Added `appsecEnabled` to TraceConfig, resolved from `DD_APPSEC_ENABLED` in `getConfig()` like every other flag.

src/trace/listener.ts
```diff
@@ -213,6 +233,10 @@ export class TraceListener {
       // Always clear the tree to prevent memory leaks, even if we skip span creation
       clearTraceTree();
     }
+    const responseStatusCode = result?.statusCode?.toString();
+    const responseHeaders = result?.headers as Record<string, string> | undefined;
+    processAppsecResponse(this.tracerWrapper.currentSpan, responseStatusCode, responseHeaders);
```
purple4reina: Can we move the is appsec enabled check to wrap this new block? That way none of it will execute if appsec isn't enabled.  It will also make it easier to see on a quick read that this section is skipped in most cases.

src/trace/listener.ts
```diff
@@ -115,6 +122,8 @@ export class TraceListener {
   }
 
   public async onStartInvocation(event: any, context: Context) {
+    initAppsec(this.config.appsecEnabled);
```
purple4reina: This doesn't seem necessary anymore now that we've moved the bool to the config.  We should just use `this.config.appsecEnabled` around the `processAppsecRequest` and `processAppsecResponse` methods directly.

src/trace/listener.ts
purple4reina: nit. I would make the signature of this method like

```js
processAppsecResponse(this.tracerWrapper.currentSpan, result)
```

Do the extraction of the status code and headers inside the function, rather than before it.  That keeps this already long onEndInvocation method neater and puts the extraction of the data closer to where you'll use it.

src/appsec/index.ts
```diff
@@ -0,0 +1,47 @@
...
+import { extractHTTPDataFromEvent } from "./event-data-extractor";
+
+const startInvocationChannel = dc.channel("datadog:lambda:start-invocation");
+const endInvocationChannel = dc.channel("datadog:lambda:end-invocation");
+
+let enabled = false;
+
+export function initAppsec(appsecEnabled: boolean): void {
+  enabled = appsecEnabled;
+}
```
purple4reina: I think these lines can be removed as long as we do the appsec enabled check just before each process call.

purple4reina (commented):
Just a couple of small nitpicks. Otherwise, everything looks great.

purple4reina (approved):

purple4reina:
Next steps to merge:

1. We'll need to increase the allowed binary size check since we're now bundling `@datadog/native-appsec`. We'll need to update the numbers here: https://github.com/DataDog/datadog-lambda-js/blob/621e9a1f292d2d167b6625bb456fe19a251fdd46/scripts/check_layer_size.sh#L11-L12

2. Once those tests are fixed, the e2e tests will run.  We have tests already that check for appsec when `DD_APPSEC_ENABLED=true`.  Currently, nodejs tests are marked as xfail, so this change should make them start passing.  We have the tests configured to alert us when a test unexpectedly passes.  So, we would expect the nodejs appsec in tracer tests to be listed as failures.  If that makes any sense.  I say just ping me when the e2e tests have finished running, before you merge, and I'll take a look.

PR #762, opened 2026-04-13
fix: [SVLS-8538] apply durable function tags via setTag on the active span

lym953:
## Problem
The tags `aws_lambda.durable_function.execution_name` and `aws_lambda.durable_function.execution_id` do not appear on the `aws.lambda` span as expected.

## Summary

- The tags were being passed through `options.tags` in `tracer.wrap()`, but that path does not reliably propagate to the actual span
- Fix: apply them directly via `span.setTag()` in `onEndingInvocation()`, the same mechanism used for `http.status_code`
- Added a test verifying `setTag` is called with the durable function context values

## Test
**Steps:**
- Build a JS layer
- Install it on a function
- Invoke the function

**Result:**
Tags appear on the `aws.lambda` span
image

### DIFF
--- a/src/trace/listener.spec.ts
+++ b/src/trace/listener.spec.ts
@@ -544,4 +544,28 @@ describe("TraceListener", () => {
       );
     });
   });
+
+  it("sets durable function context tags directly on the aws.lambda span", async () => {
+    const mockSetTag = jest.fn();
+    const mockSpan = { setTag: mockSetTag };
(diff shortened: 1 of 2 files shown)


duncanista (approved):

lym953:
The job "[dd-gitlab/publish layer sandbox" keeps failing. Ignoring it and merging this PR.

PR #784, opened 2026-06-09
fix(handler)!: drop dist/handler.js to support ESM Lambda handlers

### LINKED ISSUE #782: ESM handler resolution
## Expected Behavior

`dist/handler.js` should not be present in the published package for Node.js runtimes ≥ 14, as it causes the Lambda bootstrap to resolve the CJS entry point instead of `dist/handler.mjs`, preventing ESM user modules from loading correctly.

## Actual Behavior

`dist/handler.js` is present in the published dist because `tsc` compiles `src/handler.ts` to `dist/handler.js` and it is never removed by the post-build script (`scripts/update_dist_version.sh`). When the Lambda bootstrap resolves the handler string `node_modules/datadog-lambda-js/dist/handler.handler`, it finds `handler.js` first and loads it as CJS. This causes `loadSync` → `_tryRequireSync` to be invoked, which does not attempt to load `.mjs` files, resulting in a failure when the user's module is ESM.

The root cause is in `update_dist_version.sh`:
```sh
cp src/handler.* dist/
```
This copies `src/handler.mjs` and `src/handler.cjs` into `dist/`, but `dist/handler.js`  produced by `tsc` from `src/handler.ts`, is never cleaned up.

## Steps to Reproduce the Problem

1. Create a Lambda function with an ESM bundle (e.g. `index.mjs`) and a `package.json` with `"type": "module"` at the task root.
2. Set the Lambda handler string to `node_modules/datadog-lambda-js/dist/handler.handler` (standard Datadog handler redirect).
3. Deploy and invoke the Lambda on Node.js 18 or above.

## Specifications

- Datadog NPM version: 12.138.0 (latest at time of writing `dist/handler.js` [confirmed present](https://www.npmjs.com/package/datadog-lambda-js?activeTab=code))
- Node version: 18.x and above (also reproduced on 24.x)

## Stacktrace

```
{
  "errorType": "Error",
  "errorMessage": "require() of ES Module /var/task/index.mjs from /var/task/node_modules/datadog-lambda-js/dist/runtime/user-function.js not supported.",
  "stack": [
    "Error [ERR_REQUIRE_ESM]: require() of ES Module /var/task/index.mjs ...",
    "    at _tryRequireSync (.../datadog-lambda-js/dist/runtime/user-function.js)",
    "    at _loadUserAppSync (.../datadog-lambda-js/dist/runtime/user-function.js)",
    "    at loadSync (.../datadog-lambda-js/dist/runtime/user-function.js)",
    "    at Object.<anonymous> (.../datadog-lambda-js/dist/handler.js)"
  ]
}
```

**Workaround:** Manually delete `dist/handler.js` after `npm install` in the Dockerfile (as documented in #295). This was flagged in 2022 with a promise to remove the file after Node 12 EOL, but it remains in the package today.

**Suggested fix:** Add `rm dist/handler.js` to `update_dist_version.sh` after the `cp src/handler.* dist/` step, since the `.mjs` and `.cjs` variants now fully cover all supported runtimes.

zarirhamza:
### What

Fixes #782. Removes `dist/handler.js` from the published npm tarball; Lambda's resolver falls through to `dist/handler.mjs`, whose async `load()` handles both CJS and ESM user modules. Aligns the npm package with the layer (which has only ever shipped `handler.mjs`).

### Breaking change

Container-image deployments require `aws-lambda-ric >= 3.0.0` (June 2023). AWS-published `public.ecr.aws/lambda/nodejs:N` base images include it since mid-2023; rebuild stale custom images if needed. Older RIC versions surface as `Cannot find module 'handler'`. Historical context: `dist/handler.js` was originally added in #307 to work around exactly this in early RIC.

### Why not the shim (earlier commits in this PR)

Initial commits added a CJS shim that detected ESM at runtime and dynamically imported `handler.mjs`. Removed after review: the shim defers tracer + user-handler init from Lambda's init phase into the first invocation, breaking Provisioned Concurrency and SnapStart. `handler.mjs`'s top-level `await` pre-pays that cost. Aligns with #697 and the reporter's suggested fix in #782.

### Tests

- Existing `esm_node{18,20,22,24}` tests (layer path) unchanged
- New `container-{cjs,esm}_node{18,20,22,24}` tests exercise `dist/handler.handler` through AWS-stock RIC

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


zarirhamza:
First push regressed the `esm_node*` integration tests on all four runtimes — error was `Runtime.ImportModuleError: Cannot find module 'esm'` through `handler.cjs`. Root cause: the `Dockerfile` copies `dist/` wholesale into the layer, so the new `dist/handler.js` shim ended up on the layer too, and Lambda's bootstrap resolved `handler.handler` to it instead of falling through to `handler.mjs`.

Inside the layer, the shim's `userIsEsm()` check returns false for the integration test's `esm_node` function (handler string is `esm.handle` with no `.mjs` suffix, and `integration_tests/package.json` has no `type: "module"`), so it delegated to `handler.cjs` → `loadSync` → `_tryRequireSync`, which doesn't try `.mjs`.

The layer has never shipped a `handler.js` (it was only injected into the npm tarball by the publish-script `cp`), and Lambda's resolver fall-through to `handler.mjs` already handles both CJS and ESM via the async `load()`. Second commit (`270a846`) drops the shim from the layer via `RUN rm` so the layer keeps that behavior unchanged from main. Shim still ships in the npm package, which is where it's actually needed.

duncanista (commented):
Thanks for tackling this — the bug is real and the runtime-branching shim is a reasonable approach. A few thoughts on the structure and on coverage that I think are worth addressing before merge. Nothing blocking the direction, just some cleanup that would make the file layout easier to reason about for the next person who touches it.

src/handler.js
```diff
@@ -0,0 +1,63 @@
...
+//     fixes the ERR_REQUIRE_ESM failure mode reported in
+//     https://github.com/DataDog/datadog-lambda-js/issues/782.
+
+"use strict";
+
+const fs = require("fs");
+const path = require("path");
+
+function userIsEsm() {
+  const handlerEnv = process.env.DD_LAMBDA_HANDLER || process.env._HANDLER || "";
```
duncanista: Small naming nit: `userIsEsm()` reads like a claim about the user's code, but what this really answers is "should we take the ESM-routing branch?" — it's a heuristic routing decision (package.json sniff + env regex), not an authoritative fact. Something like `shouldUseEsmHandler()` would make the call site read more honestly and signal that it's a decision, not a property.

src/handler.js
```diff
@@ -0,0 +1,63 @@
...
+        cachedHandler = mod.handler;
+      } catch (error) {
+        cachedError = error;
+        throw error;
+      }
+    }
+    return cachedHandler(event, context);
+  };
+} else {
+  module.exports = require("./handler.cjs");
```
duncanista: Worth flagging that after this change, `src/handler.cjs` is no longer reachable as a Lambda entry point — Lambda's resolver only looks at `.js`/`.mjs`, so the only thing that ever loads `handler.cjs` is this one `require` call inside the same package. It effectively becomes a private implementation file that exists solely to give the shim something to delegate to.

That's a little awkward structurally — future readers will grep for "what loads handler.cjs?" and find one caller in the same directory. Two options worth considering:

1. Inline the CJS branch directly here (the body of `handler.cjs` is ~30 lines of `loadSync` + `initTracer` + `exports.handler = datadog(...)`), and delete `src/handler.cjs`. End state: `handler.js` = CJS entry + CJS impl, `handler.mjs` = ESM entry + ESM impl. Two files, each loaded by something.
2. Keep `handler.cjs` but add a header comment explaining that it's no longer a Lambda entry point and only exists as the delegate target.

Option 1 is my preference — minimizes the file count and matches the symmetry of `handler.mjs` (which is both an entry point on the layer and the delegate target for the ESM branch here). Happy to defer if you'd rather keep the diff tight, but at least a comment in `handler.cjs` explaining the new role would help.

src/handler.js
```diff
@@ -0,0 +1,63 @@
+// CJS entry shim used by AWS Lambda's bootstrap when the handler string
```
duncanista: Could we get a unit test for the detection logic? Even a small `src/handler.spec.js` (or wherever fits the existing test layout) covering the four cases:

- `_HANDLER=foo.mjs.handler` → ESM branch
- `LAMBDA_TASK_ROOT` points at a `package.json` with `type: "module"` → ESM branch
- `package.json` exists without `type: "module"` → CJS branch
- `package.json` missing/unreadable → CJS branch (matches prior behavior)

The local smoke tests on Node 24 mentioned in the description are great for confidence today, but module-resolution code tends to be the kind of thing that silently regresses on a future Node bump or a refactor unless there's a test guarding it. Detection here is also the load-bearing piece — if it gets the wrong answer, ESM users either keep seeing `ERR_REQUIRE_ESM` (false negative) or CJS users pay an unnecessary async-import cold-start cost (false positive). Cheap to lock in with a few mocked-fs tests.

Dockerfile
```diff
@@ -13,6 +13,13 @@ RUN yarn install
 # Build the lambda layer
 RUN yarn build
 RUN cp -r dist /nodejs/node_modules/datadog-lambda-js
+# The CJS shim at dist/handler.js exists to route ESM Lambdas through
+# handler.mjs when the npm package is used directly (the publish scripts no
+# longer copy handler.cjs to handler.js). In the layer, Lambda's bootstrap
+# resolves `handler.handler` to handler.mjs as long as handler.js is absent,
```
duncanista: The comment here is really helpful, thanks for writing it out. Minor: the relationship between `update_dist_version.sh`'s `cp src/handler.* dist/` (which is what ships the new shim) and this `rm` is implicit — if someone later renames or scopes that copy more narrowly, this `rm` could silently start failing or, worse, succeed against the wrong file. Not asking for a code change, but maybe one extra line in the comment pointing at `scripts/update_dist_version.sh` as the source of the file being removed?

src/handler.js
zarirhamza: Renamed in `f301d30`. The call site now reads `if (shouldUseEsmHandler())`, which signals it's a routing decision rather than a claim about the user's code.

src/handler.js
zarirhamza: Went with Option 1 in `f301d30`. Inlined the `handler.cjs` body into the CJS branch of `src/handler.js` and deleted `src/handler.cjs`. The heavy `require()`s (`./index.js`, `./utils/index.js`, `./runtime/index.js`, `./runtime/module_importer`) are scoped to inside the CJS branch so ESM users don't pay the cost of loading the full tracer at require-time when they only need `handler.mjs`. End state matches the symmetry you described: `handler.js` is the CJS entry + CJS impl, `handler.mjs` is the ESM entry + ESM impl.

While doing this I also noticed `scripts/update_dist_version.sh`'s `cp src/handler.* dist/` glob was now silently picking up the new `src/handler.spec.ts` — would have shipped a test fixture in the npm tarball and the layer. Tightened the `cp` to list `src/handler.js` and `src/handler.mjs` explicitly.

src/handler.js
zarirhamza: Added `src/handler.spec.ts` in `f301d30` covering the four cases you listed plus two extras:

- `_HANDLER` ends in `.mjs` → ESM
- `DD_LAMBDA_HANDLER` ends in `.mjs` → ESM
- `LAMBDA_TASK_ROOT/package.json` has `type: "module"` → ESM
- `package.json` exists without a `type` field → CJS
- `package.json` is missing → CJS
- `package.json` is unparseable → CJS

`shouldUseEsmHandler` is exposed as `_shouldUseEsmHandler` (test-only, underscored) so the test can call it directly without going through the module's branching logic. The test forces the first `require("./handler")` through the ESM branch by setting `DD_LAMBDA_HANDLER` to a `.mjs` value before requiring — that branch is side-effect-free (just defines an async lazy-import function), so the module loads in the test process without trying to `loadSync` a real user app. Detection reads `process.env` / `fs` at call time, so each test mutates the environment freely after the initial require. Uses real `fs.mkdtempSync` temp dirs rather than mocking `fs`, since the detection function only does one `readFileSync` and the temp-dir approach is faster than wiring a mock.

All 6 tests pass locally on Node 24.

Dockerfile
zarirhamza: Added a paragraph in `f301d30` pointing at `scripts/update_dist_version.sh` as the source of the file being removed, so any future change to how `src/handler.js` ends up in `dist/` will visibly bump up against this `rm` step. Also tightened the `cp` in `update_dist_version.sh` itself (from the broad `src/handler.*` glob to an explicit `src/handler.js src/handler.mjs`), which makes the contract between that script and this `RUN rm` a lot less surprising — there is now exactly one named file being copied in and removed back out.

DarcyRaynerDD:
I think I wrote the original version of the handler.cjs/esm module split a few years ago, so maybe this context would help.

So, much of this logic was due to whether the node runtime supports ESM modules and top level await, (at the time we supported node 12 which didn't). Top level await is only supported in ESM, and is needed when dynamic importing ESM modules so we can receive the handler function instead of a promise, and export the result. This isn't possible in CJS, while you can import ESM modules dynamically, they end up as promises and if you want to await that promise you have to do it within an async function. ESM doesn't have any problems requiring CJS modules dynamically though.  So, when the customer is using the layer, and a runtime >= 14, it makes sense to just use the ESM handler since it should handle all cases in theory. Since older runtimes are completely deprecated now, we should only bundle the ESM handler into the layer. I haven't looked at the support case to check if this reasoning was solid, just what I remember verifying at the time, and I know that was the previous attempted approach, but I'm confused how it could have broken CJS users.

The second case is when installing datadog-lambda-js via npm. Since we don't know which module format the customer is using, we have to package both versions. I believe I did some experimentation to find which files would be auto-imported based on their file extension. I think you had to specify the .mjs extension for it to work for esm. Typically though, a customer using the npm module could just import the datadog-lambda-js library and wrap manually if they ran into an issue.

src/handler.js
zarirhamza: Update: in commit `ee53944` I went further than inline — dropped the shim entirely and made `handler.mjs` the sole Lambda entry point in both the npm tarball and the layer. End state is one file, not two.

The trigger was [Darcy's comment](https://github.com/DataDog/datadog-lambda-js/pull/784#issuecomment-4681815700) and a closer look at the cold-start path. The shim's CJS branch saved ~10-30ms of init time on cold-start (negligible vs. the 500ms-3s a tracer-instrumented Lambda already spends in init), but its ESM branch moved 300-800ms of tracer + user-handler load work from Lambda's init phase into the **first invocation**. That's worse in three concrete ways:

1. First-invoke timeout risk — that load work now eats into the invocation's time budget (3s API Gateway is common).
2. Provisioned concurrency is wasted — work deferred past init isn't pre-warmed.
3. Lambda SnapStart is wasted for the same reason — snapshot is taken after init, and the shim's lazy `import()` runs post-snapshot.

`handler.mjs`'s top-level `await load(...)` runs under Lambda's init phase, so PC / SnapStart customers pre-pay the cost and get fast invocations. CJS users on the npm-redirect path now go through that same path; the ~10-30ms init delta is the only thing they pay, in exchange for the bug fix landing for ESM users.

This is also what the customer originally suggested in [#782](https://github.com/DataDog/datadog-lambda-js/issues/782), what SLES-2888's TAM internally verified as the working workaround, and the intent of the earlier [PR #697](https://github.com/DataDog/datadog-lambda-js/pull/697).

PR title and description updated to reflect the new shape. Re-review welcome — diff against `main` is now ~6 lines of source code changes plus removal of the old files.

zarirhamza:
@DarcyRaynerDD thanks for the historical context — it's exactly what was missing. You're right on both counts.

I switched the PR to do what you suggested in commit `ee53944`: `handler.mjs` is now the sole Lambda entry point in both the npm tarball and the layer. The shim is gone, `src/handler.cjs` is gone, and `dist/` ships a single handler file. The diff against `main` collapses down to ~6 source lines plus removal of the dead files.

Re your "I'm confused how it could have broken CJS users" point — I went back through this carefully and I don't think it would have. Lambda's bootstrap resolves `dist/handler.handler` to `handler.mjs` once `.js` is absent, and `handler.mjs`'s async `load()` already handles both CJS and ESM user modules transparently. The earlier attempt (PR #697) seems to have been closed without a recorded technical reason; I suspect it was process drift rather than a real failure. The integration test suite's `esm_node{18,20,22,24}` jobs cover exactly this code path against a real ESM user function, and they pass on this branch.

I had originally added a runtime-branching shim to avoid making CJS users on the npm-redirect path pay an extra ~10-30ms of init time (sync `loadSync` vs. async `await load()`). On reflection that's noise relative to a Datadog Lambda's normal init cost, and the shim's deferred-load behavior actually penalized ESM users on first invocation by 300-800ms — including PC and SnapStart customers, where init work is supposed to be pre-paid. Net loss for the customers this PR exists to help, so it's gone.

Description updated. Welcome any further notes.

lucaspimentel:
According to 🤖 

**`loadSync` is now dead code after this PR**

Deleting `src/handler.cjs` removes the only consumer of `loadSync`, leaving an orphaned synchronous loader chain in `src/runtime/user-function.ts`:

- `loadSync` (line 317) — its only callers were `handler.cjs` (the `traceExtractor` load and the wrapped-handler load); both are gone with the file.
- `_loadUserAppSync` (line 224) — only called by `loadSync`.
- `_tryRequireSync` (line 164) — only called by `_loadUserAppSync`.

It's also not reachable through the package's public entry point: `src/index.ts` imports only `subscribeToDC` from `./runtime`, so `loadSync` was never exposed to external consumers despite the `export { load, loadSync }` re-export in `src/runtime/index.ts:1`. No tests reference it either.

Since this PR is what orphaned it, it'd be clean to remove the three functions (`user-function.ts:164-196`, `224-242`, `317-342`) and drop `loadSync` from the `runtime/index.ts` re-export in the same change. The async `load` / `_loadUserApp` / `_tryRequire` path remains the sole live loader, which matches the PR's goal of a single ESM entry point.

scripts/update_dist_version.sh
```diff
@@ -15,7 +15,11 @@ echo "$MAIN_CONSTANTS" |
...
   sed "s/\(ddtraceVersion =\) \"\(X\.X\.X\)\"/\1 \"$DD_TRACE_VERSION\"/" > ./dist/trace/constants.js
 
-echo "Copying handler js files"
-cp src/handler.* dist/
+echo "Copying handler files"
+# Only handler.mjs ships as a Lambda entry point. Lambda's bootstrap resolves
+# `dist/handler.handler` to handler.mjs (it falls through `.js` -> `.mjs`),
+# and handler.mjs's async `load()` handles both CJS and ESM user modules, so
+# a separate `.js` / `.cjs` variant is no longer needed.
+cp src/handler.mjs dist/
```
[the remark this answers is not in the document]
zarirhamza: Good call. Added `rm -f dist/handler.js dist/handler.cjs` to `scripts/update_dist_version.sh` in `3aa51e2`, right before the `cp src/handler.mjs dist/` step. The publish scripts (`.gitlab/scripts/publish_npm.sh`, `scripts/publish_prod.sh`) already do `rm -rf ./dist` before `yarn build`, so this is purely a guard for local dev builds where stale artifacts from a previous ref could otherwise survive into the next build.

zarirhamza:
@lucaspimentel good catch — verified and addressed in `3aa51e2`:

- `loadSync` only ever had two callers, both in `handler.cjs` (the `traceExtractor` load and the wrapped-handler load). Gone.
- `_loadUserAppSync` only called by `loadSync`. Gone.
- `_tryRequireSync` only called by `_loadUserAppSync`. Gone.
- Confirmed not in the public API: `src/index.ts` only imports `subscribeToDC` from `./runtime`, and no tests reference `loadSync`.

Removed all three functions from `src/runtime/user-function.ts` and dropped `loadSync` from the `src/runtime/index.ts` re-export in the same chunk. Build + lint + tests pass; the two `nock.isDone()` flakes in `src/index.spec.ts` reproduce on plain `origin/main` and are unrelated.

Net diff: -105 lines from `user-function.ts`, -1 char from `runtime/index.ts`.

joeyzhao2018 (approved):
👍 :shipit:

duncanista (approved):

rithikanarayan:
Merging with unrelated e2e failures (dotnet related).

PR #791, opened 2026-06-18
test(integration): drop Node 18 from integration matrix

duncanista:
## Summary

Drops \`nodejs18.x\` from the integration test matrix. CI-only change — no customer-facing behavior.

## Why

AWS deprecated \`nodejs18.x\` on **Sep 1, 2025**:
- Block-create: Feb 1, 2027
- Block-update: Mar 3, 2027

New ESM/Lambda development is on Node 20/22/24. Keeping Node 18 in CI burned ~25% of every integration-test run on a runtime AWS is winding down.

## Changes

- \`scripts/run_integration_tests.sh\`: remove the \`node18=(...)\` declaration and drop \`"node18"\` from \`PARAMETERS_SETS\`
- Delete 70 orphaned Node 18 snapshot files (63 return-value + 7 log) so the snapshot directory stops accumulating dead fixtures

Scope intentionally narrow — layer builds, Dockerfile, GitLab pipelines are untouched; those have separate deprecation timelines.

## Test plan

- [ ] Integration test run on a sandbox account succeeds for Node 20/22/24 with no Node 18 deploy attempts


duncanista:
Closing — dropping Node 18 from the integration test matrix is premature while the library still builds and publishes Node 18 layers (`scripts/build_layers.sh`) and the npm package still installs against Node 18 functions. AWS deprecation alone isn't sufficient justification; coverage gap risk is real (a change could break Node 18 customers undetected).

Will revisit as part of a coordinated Node 18 EOL across the library (stop layer publication + README announcement + drop from CI together).

PR #804, opened 2026-07-29
test: add local Docker/RIE integration test harness (Level 2 tests)

joeyzhao2018:
## What this is

This is the first PR in the effort to move `datadog-lambda-js`'s core logic into `dd-trace-js`.
It doesn't move any code yet — it builds the safety net that every later PR will be checked against.

Per Step 2 of the [migration strategy](https://docs.google.com/document/d/1KeEmUIfpUj7h4fO_iyIuGOvNC92nWqp-O09NulOkdrg/edit?tab=t.et8e66pgxbe7), we need three levels of test coverage. Level 2 — Docker-based integration tests — didn't exist yet: our integration tests run against real AWS Lambdas, which is too slow and too credential-bound for the dd-trace pipeline's few-minute budget. This PR adds that missing level.

## What it does

Adds `integration_tests_local/`, which runs the container handler variants against the AWS [Runtime Interface Emulator](https://github.com/aws/aws-lambda-runtime-interface-emulator) — no AWS account, no deployed lambda functions, no credentials.

For each of Node 18/20/22/24 × {cjs, esm}, it builds the image, starts it under RIE, fires 9 input events, and diffs both the return values and the normalized logs against committed snapshots. ~~That's **80 golden snapshots** (72 return values + 8 log files) capturing exactly what the library does today.~~ We actually enhance this part to remove replications in the snapshots. so now it's much cleaner.

Those goldens are the point. Once we start moving code into `dd-trace-js`, they're the oracle that tells us whether migrated code still behaves identically — so they're captured from the current implementation *before* any migration happens, and they're never regenerated wholesale from migrated code.

Also adds a GitHub Actions workflow running all five runtime legs on every PR.

## Notes for review

~~Most of the diff is the 80 snapshot files.~~ The parts worth actual review are `run.sh` and `normalize.sh`.

**Normalization is the load-bearing piece.** It decides which volatility gets erased before diffing (timestamps, request IDs, durations, trace IDs). Erase too little and the tests flake; erase too much and the gate silently stops catching regressions. Two specific decisions:

- Span `meta`/`metrics` keys are **preserved** — their stable keys and values are part of the compatibility contract.
- Trailing-whitespace stripping is scoped to RIE's `REPORT` lines only, so application output  stays an exact-whitespace comparison.

**Node 18 and 20 legs are deliberate.** dd-trace's `v5.x` is the only line serving those runtimes, and we ship layers for them, so they stay in the matrix for the whole migration.
(This is the same position taken when closing #791.)

**Node 26 is a special case.** `public.ecr.aws/lambda/nodejs:26` doesn't exist yet — AWS only publishes `26-preview.*` — so the base image is pinned to a specific preview tag, threaded through a runtime→image-tag indirection applied only at the `docker build --build-arg` (the bare major still drives image tags, container names and snapshot paths). Node 26 has no committed snapshots yet, so it's excluded from the default local run and runs in update mode
in CI, which asserts it produces exactly the expected 20 artifacts without committing them. The next PR freezes them. When AWS publishes the GA image, the Node 26 goldens need re-capturing from the pre-migration code — that's recorded in the README.

Hardening details: RIE binary pinned to v1.36 with a per-platform SHA-256 that's re-verified on every run; missing snapshots fail instead of passing silently; HTTP status is checked alongside the response body (`curl -s` exits 0 on a 5xx, so update mode could otherwise record error pages as goldens); `diff` exit codes 0/1/>1 are handled distinctly; fixed sleeps replaced with polling on both `REPORT` and `INVOKE RTDONE`, since RIE can emit them out of order.

## Trying it locally

```bash
./integration_tests_local/run.sh                          # all runtimes with committed snapshots
RUNTIME_PARAM=18 VARIANT_PARAM=esm ./integration_tests_local/run.sh
UPDATE_SNAPSHOTS=true ./integration_tests_local/run.sh    # regenerate
```

Requires Docker. First run pulls the base images (~1 GB each) and the RIE binary.

## What comes next

This is Step 2 of the strategy. The PRs that follow, roughly in order:

1. **Freeze the baseline** — capture the Node 26 snapshots and tag the full 100-snapshot baseline, making all five runtime legs strict comparisons.
2. **dd-trace-js scaffolding** — register an `aws-lambda` plugin and define a small, stable facade (`wrap`, `sendDistributionMetric`, `sendDistributionMetricWithDate`,  `getTraceHeaders`) for the shim to call.
3. **Mechanical ports** (Step 3) — utilities and event detection, context extractors, span inferrer and X-Ray, span pointers. Code moves as-is; specs move with it.
4. **The seams** — the places copy-paste can't solve: config wiring, handler lifecycle, metrics, and cold start / DSM / AppSec.
5. **Shim conversion** (Step 4) — `datadog-lambda-js` becomes thin and delegates to dd-trace, with the layer built from migrated logic.
6. **Release gates** — goldens, ported unit tests and the real-AWS e2e suite all run against the candidate layer before anything is deleted.
7. **Cleanup and release** (Step 5) — delete the old business logic, publish.

Every one of those is checked against the goldens this PR adds, and progress is tracked row-by-row in `migration_parity.md`.


integration_tests_local/normalize.sh
```diff
@@ -0,0 +1,82 @@
...
+local_dir=$(dirname "$script_path")
+repo_dir=$(dirname "$local_dir")
+
+# An empty Perl program is a pass-through when RUN_ID is unset.
+run_id_filter=''
+if [ -n "$RUN_ID" ]; then
+    run_id_filter="s/${RUN_ID}/XXXX/g"
+fi
+
+node "$repo_dir/integration_tests/parse-json.js" |
```
rithikanarayan, line 32: I've found in a couple of the layers (i think node and ruby) that when I run integration tests locally sometimes I get cold start logs and sometimes I don't, so olivier and I have removed them ([example](https://github.com/DataDog/datadog-lambda-js/blame/2d3f33d5bfc1d9cecdb5e5fc01d6cd82f5b27f03/scripts/run_integration_tests.sh#L246)) in these repos. You might want to consider doing the same here, or just leave it unless it becomes a problem.

rithikanarayan: although now that I read further not sure if we can do that given the proactive init simulation

joeyzhao2018: When you previously run integration tests "locally", it actually still uses real lambdas. For real lambda, the platform decides when to run init, so sometimes it proactively initializes the sandbox and cold_start flips to false. 
This PR is real 'local' using docker and no need for aws credentials whatsoever. And under RIE there's no platform scheduler — we start the container and drive all 9 invocations ourselves, so the cold→warm transition is deterministic. 

As for the  SIMULATE_PROACTIVE_INIT mode. It is a manual diagnostic, not part of the test gate. In other words, it's not used (for now). To make it more clear, i updated the README for its section.

rithikanarayan (approved):
lgtm

PR #814, opened 2026-08-26
feat: use DD_SERVICE for inferred spans when integration service names are removed

zarirhamza:
## Summary

- When `DD_TRACE_REMOVE_INTEGRATION_SERVICE_NAMES_ENABLED=true` **and** `DD_SERVICE` is set, inferred (synthetic) event-source spans use the base service name (`DD_SERVICE`) instead of the AWS resource/instance representation.
- An explicit `DD_SERVICE_MAPPING` entry still takes precedence over `DD_SERVICE`.
- Default behavior (flag off) is unchanged, so existing integration snapshots are unaffected.

## Resolution order in `determineServiceName`

1. `DD_SERVICE_MAPPING` (specific key, then generic key)
2. If `DD_TRACE_REMOVE_INTEGRATION_SERVICE_NAMES_ENABLED` and `DD_SERVICE` set → `DD_SERVICE`
3. If `DD_TRACE_AWS_SERVICE_REPRESENTATION_ENABLED` is off → fallback
4. Extracted resource/instance name, else fallback

## Motivation

Customers currently need one `DD_SERVICE_MAPPING` entry per trigger type per function to get their event-source spans onto their own service name. This gives them a single setting that consolidates all of them at once.

Tracked by [FRSLES-869](https://datadoghq.atlassian.net/browse/FRSLES-869). Mirrors the resolution order in the Python layer ([datadog-lambda-python#834](https://github.com/DataDog/datadog-lambda-python/pull/834)) and in the extension.

## Scope

Service-name resolution only. `peer.service` on inferred spans continues to report `DD_SERVICE` as it does today; revisiting that pairing is deliberately left to a follow-up.

## Test plan

- [x] 11 new cases in `span-inferrer.spec.ts` covering all 8 trigger types, `DD_SERVICE_MAPPING` precedence, unset `DD_SERVICE`, and precedence over `DD_TRACE_AWS_SERVICE_REPRESENTATION_ENABLED=false`
- [x] `npx jest src/trace/span-inferrer.spec.ts` — 105 passed
- [x] Full unit suite — no new failures (2 pre-existing `index.spec.ts` failures also present on `main`)
- [x] Integration snapshots untouched (default behavior unchanged)

[FRSLES-869]: https://datadoghq.atlassian.net/browse/FRSLES-869?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


joeyzhao2018 (approved):
lgtm

PR #817, opened 2026-08-27
fix(appsec): send the normalized status code on end-invocation

CarlesDD:
### What does this PR do?

Two fixes on the AppSec end-invocation payload:

- Reorders `onEndingInvocation` so the HTTP status code is resolved and tagged before AppSec runs, and publishes that normalized value.
- Normalizes the response headers matching what the extractor already does for request headers.

### Motivation

AppSec was publishing the raw `result.statusCode` and raw response headers, so the WAF got no status at all when the handler omitted one and response headers in a different shape from the request ones.

[APPSEC-68818](https://datadoghq.atlassian.net/browse/APPSEC-68818)

### Testing Guidelines

`src/appsec/index.spec.ts` covers the normalized value winning over the raw one and the fallback for non-HTTP triggers. `src/trace/listener.spec.ts` adds cases over real event samples: API Gateway v2 without a status, ALB with no result, streaming with no result, the tag-before-AppSec ordering, and a 500 still reaching AppSec while short-circuiting.

`src/appsec` + `src/trace`: 475 tests passing. `tsc`, `tslint` and `prettier` clean.

### Additional Notes

Shippable on its own: it fixes the status code the WAF receives today, with or without API Security. It is also a prerequisite for API Security on Lambda, where the sampling decision is taken off the span and needs `http.status_code` already set.

### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

[APPSEC-68818]: https://datadoghq.atlassian.net/browse/APPSEC-68818?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/appsec/index.ts
```diff
@@ -30,12 +30,18 @@ export function processAppsecRequest(event: any, span: any): void {
...
+ * @param statusCode Status code already normalized by the trigger layer. Falls back to the raw
+ *   `result.statusCode` when the caller has none, which happens for non-HTTP triggers.
+ */
+export function processAppsecResponse(span: any, result: any, statusCode?: string): void {
   if (!span || !endInvocationChannel.hasSubscribers) return;
 
   endInvocationChannel.publish({
     span,
-    statusCode: result?.statusCode?.toString(),
+    statusCode: statusCode ?? result?.statusCode?.toString(),
```
litianningdatadog: is it possible that statusCode has different value from results.statusCode, assuming neither is null?

litianningdatadog (commented):
left some comment

src/appsec/index.ts
CarlesDD: only for non-http triggers, but the tracer drops those end-invocation messages away, so it was dead code. removed.

src/trace/listener.ts
```diff
@@ -231,25 +231,25 @@ export class TraceListener {
       // Always clear the tree to prevent memory leaks, even if we skip span creation
       clearTraceTree();
     }
-    if (this.config.appsecEnabled) {
-      processAppsecResponse(this.tracerWrapper.currentSpan, result);
-    }
+    let statusCode: string | undefined;
```
CarlesDD, line 234: The status code has to be resolved before AppSec runs, because the WAF needs the normalized value: a handler may leave `statusCode` out of its response, and only this call knows what to nswer then.

it also has to be tagged first. The API Security sampler reads the span's priority, and when it is not set yet it triggers the trace-level priority sampler, which evaluates sampling rules against the span tags, so `http.status_code` must already be there.

src/appsec/headers.ts
```diff
@@ -0,0 +1,35 @@
...
+        result[key.toLowerCase()] = values.join(", ");
+      }
+    }
+  }
+
+  if (headers) {
+    for (const [key, value] of Object.entries(headers)) {
+      const lowerKey = key.toLowerCase();
+      if (!(lowerKey in result) && value !== undefined) {
+        result[lowerKey] = value;
```
litianningdatadog: if we plan to keep this normalizeHeaders() generic, you may want to expand header value to support input type such as number and bool and stringfy in result?

litianningdatadog (approved):
Stamped. Left a non-blocker comment

PR #819, opened 2026-08-28
perf(runtime): use dd-trace loader registration

BridgeAR:
Load the function-local `dd-trace` registration entry during tracer initialization instead of registering another loader hook from the Lambda wrapper.

Direct `Module.register()` creates a loader worker and synchronous startup communication. The canonical entry can select in-process hooks. Older function-local tracers without `register.js` retain the direct fallback.

Paired Lambda trials reduced module load by 79.549 ms and startup total by 65.788 ms (21.47%). A fresh local Node 22.23.2 benchmark across 40 alternating pairs reduced module load by 45.881 ms and startup total by 45.264 ms (15.54%).


joeyzhao2018 (approved):
LGTM

joeyzhao2018 (approved):
LGTM! Thank you

PR #824, opened 2026-09-01
ci(integration): stabilize Lambda log snapshots

BridgeAR:
CloudWatch returned successful responses before every invocation log was visible. The runner accepted those partial results as snapshot input.

AWS can initialize a sandbox more than ten seconds before its first invocation. This adds proactive-initialization telemetry and changes `cold_start`, but the broad snapshots treated both scheduler-owned fields as library regressions. Stored zip snapshots also kept a blank line from the former `INIT_START` substitution.

The runner now polls until every expected completion record is present and uses one AWS/RIE-aware normalizer for current output and snapshots. AWS removes scheduler-owned fields and historical padding. RIE retains `cold_start`, and focused trace and metric tests pin proactive-initialization behavior.

The runner also resolves its helper directory before changing directories and keeps successful Serverless diagnostics on stderr. Polling starts immediately instead of waiting for the former fixed 20-second delay.


joeyzhao2018 (approved):
LGTM. good catch. btw, we are also migrating these real aws-lambda based tests to serverless-e2e-tests

PR #830, opened 2026-09-08
feat: extract trace context from MSK events

### LINKED ISSUE #829: Built-in trace context extraction for MSK-triggered Lambdas
Hi! I’m running a Node.js 24 Lambda triggered by Amazon MSK, using Datadog layer 142. Messages arrive successfully from a dd-trace-instrumented producer, but the Lambda trace isn’t connected to the producer trace.

The incoming records contain Datadog trace headers as byte arrays in `records[topicPartition][i].headers`. I couldn’t find built-in MSK trace context extraction in this library. I also noticed similar support was added to the Lambda extension in [#1066](https://github.com/DataDog/datadog-lambda-extension/pull/1066).

Would you welcome a PR adding this support? I’d be happy to contribute an implementation with tests and a minimal reproduction.

lucassarcanjo:
### What does this PR do?

Adds automatic trace context extraction for MSK-triggered Lambdas. Kafka header byte arrays are decoded as UTF-8 and passed to the existing tracer, preserving Datadog and W3C propagation context.

For batches, the Lambda span uses the first record with valid trace context across the topic-partition groups. Malformed headers are skipped, and headers from different records are never combined.

### Motivation

Fixes #829. An instrumented producer's trace headers reach the Lambda event, but the existing dispatcher does not extract them, so the Lambda starts a separate trace.

### Testing Guidelines

- `yarn test --runInBand`: 691 tests and 3 snapshots passed on Node.js 22.
- `yarn lint` and formatting checks passed.
- Local Node.js 24 check with real `dd-trace@5.118.0`: the Lambda tracing lifecycle preserves the producer parent ID, sampling priority, and 128-bit trace ID with Datadog-only, W3C-only, and combined headers. The same reproduction fails with the original dispatcher.
- Live AWS/MSK and Docker integration validation are pending.

### Types of Changes

- [x] New feature

### Check all that apply


src/trace/context/extractors/msk.ts
```diff
@@ -0,0 +1,45 @@
...
+          const headers = this.getParsedRecordHeaders(record);
+          if (Object.keys(headers).length === 0) continue;
+          const traceContext = this.tracerWrapper.extract(headers);
+          if (traceContext) return traceContext;
+        } catch (error) {
+          handleExtractionError(error, "MSK");
+        }
+      }
+    }
+    return null;
```
purple4reina, line 37: This all looks good to me.  The other extractors log on success/failure.  Would be nice to have that here as well.

lucassarcanjo: Thanks @purple4reina, I’ve added the logs.

src/trace/context/extractors/msk.ts
```diff
@@ -0,0 +1,50 @@
...
+          handleExtractionError(error, "MSK");
+        }
+      }
+    }
+    logDebug("Failed to extract trace context from MSK event");
+    return null;
+  }
+
+  private getParsedRecordHeaders(record: MSKRecord): Record<string, string> {
+    const headers: Record<string, string> = Object.create(null);
```
BridgeAR:
```suggestion
  private getParsedRecordHeaders(record: MSKRecord): Record<string, string> | undefined {
    const headers: Record<string, string>;
```

By doing this, there is no need to call `Object.keys().length === 0` above.

We just need to lazy assign below with `headers ??= {}`

src/trace/context/extractors/msk.ts
```diff
@@ -0,0 +1,50 @@
...
+    if (!Array.isArray(record?.headers)) return headers;
+
+    for (const entry of record.headers) {
+      if (!entry || typeof entry !== "object" || Array.isArray(entry)) continue;
+      for (const [name, value] of Object.entries(entry)) {
+        // MSK serializes Kafka header bytes as integer arrays, not base64.
+        // Validate before decoding because Buffer.from silently coerces invalid bytes.
+        if (Array.isArray(value) && value.every((byte) => Number.isInteger(byte) && byte >= 0 && byte <= 255)) {
+          headers[name.toLowerCase()] = Buffer.from(value).toString("utf8");
+        }
```
BridgeAR:
```suggestion
        if (Array.isArray(value)) {
          const lowerCaseName = name.toLowerCase()
          headers[lowerCaseName] ??= {}
          headers[lowerCaseName] = Buffer.from(value).toString("utf8");
        }
```

The validation does not seem to be necessary here. AWS types guarantee number and it would just fail later, if it would be a different header, which seems fine (which in itself is very unlikely)

src/trace/context/extractors/msk.ts
```diff
@@ -0,0 +1,50 @@
...
+import { TracerWrapper } from "../../tracer-wrapper";
+import { SpanContextWrapper } from "../../span-context-wrapper";
+
+export class MSKEventTraceExtractor implements EventTraceExtractor {
+  constructor(private tracerWrapper: TracerWrapper) {}
+
+  extract(event: MSKEvent): SpanContextWrapper | null {
+    // A Lambda span can have only one parent. Use the first record with valid
+    // trace context, without combining headers from different records.
+    for (const records of Object.values(event.records ?? {})) {
```
BridgeAR: Nit: it would be nice to put this for loop into a if statement if event.records is undefined or null (is that even possible? I am not sure, while we should probably guard against a changed protocol anyway). That is faster than falling back to the empty object and calling Object.values on that.

src/trace/context/extractors/msk.ts
```diff
@@ -0,0 +1,50 @@
...
+        try {
+          const headers = this.getParsedRecordHeaders(record);
+          if (Object.keys(headers).length === 0) continue;
+          const traceContext = this.tracerWrapper.extract(headers);
+          if (traceContext) {
+            logDebug("Extracted trace context from MSK event");
+            return traceContext;
+          }
+        } catch (error) {
+          handleExtractionError(error, "MSK");
```
BridgeAR: Should we report each error? If we receive a lot of records that for what ever reason throw during parsing, this might report a lot of those, no? 😅

BridgeAR (commented):
Thank you for the PR! I just left a few suggestions to make the code a bit faster :)

src/trace/context/extractors/msk.ts
lucassarcanjo: That makes sense! Fixed

src/trace/context/extractors/msk.ts
lucassarcanjo: Yeah, I changed it to wrap the entire loop instead, matching the SQS extractor’s behavior.

src/trace/context/extractors/msk.ts
lucassarcanjo: I see. I added an early return with a debug log instead of the `?? {}` fallback.

BridgeAR (commented):
Thank you for the quick follow-ups!

Code wise it seems fine to me.

lucassarcanjo:
Hi @lym953, could you take a look at this PR?

jcstorms1:
@codex review

jcstorms1:
@codex review

jcstorms1:
LGTM 👍

jcstorms1 (approved):

zarirhamza:
Thanks @lucassarcanjo! This looks good to us but we need you to rebase onto the original main branch instead of a fork. If you'd rather us take care of it, please enable **allow edits from maintainers** so we can rebase and merge

lucassarcanjo:
Hi @zarirhamza, done!

PR #832, opened 2026-09-09
bump dd-trace v6

ojproductions:
### What does this PR do?
Bumps dd-trace v6 to most recent version. This is to include the new cold start improvements.

### Types of Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "node": ">=18"
   },
   "peerDependencies": {
-    "dd-trace": "^5.126.0 || ^6.12.0"
+    "dd-trace": "^5.126.0 || ^6.15.0"
   },
   "peerDependenciesMeta": {
     "dd-trace": {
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -44,7 +44,7 @@
     "@types/node": "^20.12.10",
     "@types/promise-retry": "^1.1.3",
     "@types/shimmer": "^1.0.1",
-    "dd-trace": "^6.12.0",
+    "dd-trace": "^6.15.0",
```
ojproductions, line 47: @joeyzhao2018 do these fixtures also need to be updated too? Or does it get overrode in the dockerfile/run scripts?

purple4reina (commented):
Can we get a better PR description and title on this?  I'm unsure why you say you're starting a release but then see you bumping the ddtrace version.  Why does this version need to be changed?

ojproductions:
@purple4reina, @joeyzhao2018 mentioned there were some cold start improvements in the most recent dd-trace release, and we wanted to get it out in this lambda release. It's not critical, we can do it in the next release.

package.json
joeyzhao2018: i think the script today only updates v5... I actually think we should also make the script update v6 too

ojproductions: ok I thought so. I'm working to remove these pins right now, so they use the root ddtrace version

rithikanarayan (approved):

PR #833, opened 2026-09-09
fix: support installs without yarn.lock

crysmags:
## Summary

- Allow the Node 18/20 dependency installer to run when its input tree does not contain `yarn.lock`.
- Preserve and restore an existing lockfile, while removing a lockfile generated during installation when the input did not include one.
- Add a regression test for the lockfile-free build context and run it in CI.

## Failure

After #818 changed the layer Dockerfile to call `scripts/install_deps.sh`, the `DataDog/dd-trace-js` default GitLab pipeline began failing while building the Node 20 Lambda layer for [dd-trace-js PR #9812](https://github.com/DataDog/dd-trace-js/pull/9812):

```text
Node 20 is not supported by dd-trace v6, pinning dd-trace to 5.126.0
cp: cannot stat 'yarn.lock': No such file or directory
```

The downstream pipeline uses a reduced build context that contains `package.json` and the installer scripts but not the repository lockfile. Yarn can install and generate a temporary lockfile in that context; only the installer's unconditional backup made the lockfile mandatory.

## Testing

- `bash -n scripts/install_deps.sh scripts/test_install_deps.sh`
- `./scripts/test_install_deps.sh`


joeyzhao2018 (approved):
Sorry and Thank you.
LGTM. :shipit:

ojproductions:
@crysmags we're currently doing a release, could you wait to merge this until after its done? shouldnt be too much longer, but will lyk if it becomes delayed

BridgeAR:
Superseded

PR #838, opened 2026-09-16
feat: extract trace context from MSK events

### LINKED ISSUE #829: Built-in trace context extraction for MSK-triggered Lambdas

jcstorms1:
### What does this PR do?

Direct copy of community PR https://github.com/DataDog/datadog-lambda-js/pull/830

Adds automatic trace context extraction for MSK-triggered Lambdas. Kafka header byte arrays are decoded as UTF-8 and passed to the existing tracer, preserving Datadog and W3C propagation context.

For batches, the Lambda span uses the first record with valid trace context across the topic-partition groups. Malformed headers are skipped, and headers from different records are never combined.

### Motivation

Fixes #829. An instrumented producer's trace headers reach the Lambda event, but the existing dispatcher does not extract them, so the Lambda starts a separate trace.

### Testing Guidelines

- `yarn test --runInBand`: 691 tests and 3 snapshots passed on Node.js 22.
- `yarn lint` and formatting checks passed.
- Local Node.js 24 check with real `dd-trace@5.118.0`: the Lambda tracing lifecycle preserves the producer parent ID, sampling priority, and 128-bit trace ID with Datadog-only, W3C-only, and combined headers. The same reproduction fails with the original dispatcher.
- Live AWS/MSK and Docker integration validation are pending; this PR is a draft.

### Types of Changes

- [x] New feature

### Check all that apply


zarirhamza (approved):

src/trace/context/extractors/msk.ts
```diff
@@ -0,0 +1,56 @@
...
+export class MSKEventTraceExtractor implements EventTraceExtractor {
+  constructor(private tracerWrapper: TracerWrapper) {}
+
+  extract(event: MSKEvent): SpanContextWrapper | null {
+    if (!event?.records) {
+      logDebug("Failed to extract trace context from MSK event");
+      return null;
+    }
+
+    try {
```
clifordshelton: The `try` wraps both loops, so one bad record aborts the scan of the entire batch.

Scenario: record 1 of a 10-record batch carries a corrupt `traceparent/tracestate` that makes `tracerWrapper.extract` throw (the spec's own "returns null after an extraction error" test locks in this behavior. Extract is called exactly once, then the loop unwinds). Records 2..N may carry perfectly valid `x-datadog-*` headers, but they're never examined, and the Lambda starts an orphan trace which is exactly the case this PR is meant to fix.

We could move the try/catch inside the inner loop so a throwing record is skipped rather than ending the scan:

```ts
for (const records of Object.values(event.records)) {
  if (!Array.isArray(records)) continue;
  for (const record of records) {
    try {
      const headers = this.getParsedRecordHeaders(record);
      if (!headers) continue;
      const traceContext = this.tracerWrapper.extract(headers);
      if (traceContext) {
        logDebug("Extracted trace context from MSK event");
        return traceContext;
      }
    } catch (error) {
      handleExtractionError(error, "MSK");
    }
  }
}
```

The "returns null after an extraction error" test would then need updating to assert the scan continues past the throwing record.

src/utils/event-validator.ts
```diff
@@ -75,6 +76,15 @@ export class EventValidator {
     return Array.isArray(event.Records) && event.Records.length > 0 && event.Records[0].kinesis !== undefined;
   }
 
+  static isMSKEvent(event: any): event is MSKEvent {
+    return (
+      event?.eventSource === "aws:kafka" &&
```
clifordshelton, line 81: Should we also check for `event?.eventSource === "SelfManagedKafka"` to cover the case of a Kafka cluster managed on EC2 or somewhere else?

clifordshelton (approved):

src/utils/event-validator.ts
clifordshelton: Checked this and it's a non-issue for this PR. MSK and self-managed Kafka are two distinct Lambda event source mapping types: an MSK ESM always emits eventSource: "aws:kafka", and a self-managed Kafka ESM always emits "SelfManagedKafka". An MSK trigger will never produce the latter, so isMSKEvent can't wrongly reject a real MSK event — the MSK path is complete as written.

Self-managed Kafka would be a scope expansion, not a gap here, and it looks like a larger piece of work than one validator line anyway (trigger.ts has no Kafka entry in eventTypes yet, so neither flavor gets trigger tags or an inferred span). Leaving it as a possible follow-up. Resolving.

